PR #14, opened 2026-04-29
feat(youtube-screenshotter): motion.py — optical-flow sprite-grid primitive

devonjones:
## Summary

Adds `scripts/motion.py` to the youtube-screenshotter skill — an optical-flow primitive that complements `discover.py` for the case it can't catch on its own: a small object moving across an otherwise-static background, where the global pHash stays within threshold and run-grouping merges the whole animation into one start frame.

The Phase 1.6 e2e against `MIqpvpNS5pI` (\"The Ox That's Breaking Your Fantasy Map\") was the canonical motivating example: at t=358 the captured frame is blank parchment (the diagram had not yet rendered), the boat slides across the river over the next ten seconds, and the only frame `discover.py` ever picked was the empty one. The synthesizer's value-filter classifier had no useful frame to classify.

`motion.py` densely re-samples a sub-range (default 4 fps), computes Farneback optical flow per consecutive pair, picks the largest moving connected component, scores each frame by `area * centeredness * edge_penalty`, and — with `--sprite-out PATH` — composes a 3x3 Brady-Bunch grid showing the whole motion arc as one image.

## What's included

- **`scripts/motion.py`** — the new entrypoint. CLI:
  ```
  motion.py <video> --start S --end E [--fps 4] [--sprite-out PATH] [--sprite-frames 9] [--sprite-cols 3]
  ```
- **`SKILL.md` description block** — extended to mention the new primitive's purpose and the trigger case it covers.
- **Caller-friendly helpers** (importable):
  - `analyse_range(...)` — main analysis entry, returns the JSON manifest.
  - `compose_sprite_grid(...)` — composites an ordered list of full-res frames into a Brady-Bunch grid.
  - `pick_grid_shape(n)` — square-ish (cols, rows) layout (4→2x2, 5-6→3x2, 7-9→3x3, …).

## Algorithm refinements during smoke-testing

Three failure modes the v1 picker had to handle, in commits separate from the initial implementation so the reasoning is auditable:

1. **Static-reveal spike at the start.** A diagram fading in produces a one-frame motion-area burst that dwarfs the sustained motion right after. Picking that frame as cell (1,1) wastes the slot on near-empty parchment. Trim when the leading sample's area exceeds 3x the median of the next three.
2. **Trailing scene-cut spike.** When the supplied range over-shoots into the next scene, the cut frame's motion-area dwarfs the recent rolling median. Compare against the rolling median of the last 5 samples (1.5x ratio) — single-frame jitter in busy animations can momentarily double per-frame area without being a cut.
3. **Multi-cluster ranges.** If the supplied range spans a quiet gap (1+ second of parchment between scenes), split into clusters and pick the largest. `CLUSTER_GAP_SAMPLES=4` (1s at fps=4) splits scene transitions but holds together within an animation's micro-pauses.

The primitive is content-agnostic by design — a talking-head with hand gestures generates motion of comparable magnitude to a sliding boat. Filtering out non-animations is the synthesizer's job (where the keep-kind has already been confirmed via vision), not motion.py's.

## Smoke-tested against MIqpvpNS5pI

Re-running the Phase 1.6 e2e with motion-bisection produces 4 sprite grids in the wiki entry where there were previously 14 single-frame keepers:

| Scene | Before | After |
|---|---|---|
| Chokepoint toll build (t=206–218) | 4 single frames | 1 × 2x2 grid |
| Surplus-lord problem build (t=281–304) | 6 single frames | 1 × 3x2 grid |
| Caloric battery (t=333–340) | 2 single frames | 1 × 3x3 motion grid |
| Boat-on-river (t=358–378) | 1 wrong frame (parchment) | 1 × 3x3 motion grid showing the whole arc |

Net: 33 keepers → 24 keepers, with each replaced section reading more clearly as one animation arc.

## Refs

bd-7h0 — Phase 1.7: motion-bisection for animated sequences.

## Test plan

- [x] `motion.py /tmp/probe-smoke.mp4 --start 0 --end 3 --fps 4` returns valid JSON manifest with no sprite (no progressive motion in synthetic 3s red video).
- [x] `motion.py MIqpvpNS5pI.mp4 --start 358.79 --end 378.25 --fps 4 --sprite-out boat-grid.png` produces a 9-frame arc spanning boat-enters → boat+cart → CARGO/EATEN bar filling red.
- [x] Trailing-cut trim correctly trims the t=382.5 sample when `--end 385` over-shoots into the talking-head scene.
- [x] `pick_grid_shape(4)` returns `(2, 2)`; `pick_grid_shape(6)` returns `(3, 2)`; `pick_grid_shape(9)` returns `(3, 3)`.
- [ ] Wiring into youtube-synthesizer Phase A.5 — deferred to a follow-up PR (the primitive is callable today; the synthesizer integration adds the auto-decision-and-compose logic).


plugins/youtube-screenshotter/skills/youtube-screenshotter/SKILL.md
```diff
@@ -60,6 +60,15 @@ scripts/extract.py "<URL_OR_VIDEO_ID>" -t 1 -t 30 -t 500 -o ./out
...
+
+```bash
+scripts/motion.py <video> --start S --end E [--fps 4] [-o DIR]
+                          [--sprite-out PATH] [--sprite-frames 9] [--sprite-cols 3]
+```
+
+The primitive is content-agnostic: a talking-head with hand gestures generates motion comparable to a sliding boat. That's fine for the grid layout — a 3×3 of slightly-different gesture poses is visually unambiguous noise the reader can ignore, while a real animation reads as a clear arc. The decision *whether* to sprite a span (vs. pick a single frame, vs. emit a text-only `> [animation: ...]` annotation) lives in the caller — typically `youtube-synthesizer` Phase A.5.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


The first two sentences here are a justification ("why this design is OK") rather than information the caller needs to use motion.py. The talking-head-vs-boat analogy and the gloss on what a 3×3 of gesture poses looks like are explanatory padding — a caller already knows what optical flow on a talking head produces. Consider collapsing to just the load-bearing sentence: "The primitive is content-agnostic; the decision *whether* to sprite a span (vs. pick a single frame, vs. emit a text-only `> [animation: ...]` annotation) lives in the caller — typically `youtube-synthesizer` Phase A.5."

devonjones: Fixed - collapsed the content-agnostic paragraph to just the load-bearing sentence about caller-side decisions. Kept the suggested wording verbatim.

plugins/youtube-screenshotter/skills/youtube-screenshotter/scripts/motion.py
```diff
@@ -0,0 +1,501 @@
...
+    cell_h = round(cell_w * src_h / src_w)
+    total_w = cell_w * cols
+    total_h = cell_h * rows
+    sprite = Image.new("RGB", (total_w, total_h), color=(0, 0, 0))
+    for i, img in enumerate(images):
+        scaled = img.resize((cell_w, cell_h), Image.LANCZOS)
+        col = i % cols
+        row = i // cols
+        sprite.paste(scaled, (col * cell_w, row * cell_h))
+    sprite.save(out_path, format="PNG", optimize=True)
```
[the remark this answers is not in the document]
devonjones: Won't fix - sprite generation is not in a hot path (one PNG per scene per video, written once and read many times in an Obsidian vault), and file size matters for git/sync performance downstream. The 1-2s save time is acceptable. Happy to revisit if a real-time use case ever shows up.

PR #18, opened 2026-05-14
bfd: ship 6 default specialist reviewer agents with C+E override semantics

devonjones:
## Summary

First cut of **devon-claude-skills-bfd**: ship 6 default specialist reviewer agents so users get useful agent reviews out-of-the-box without authoring AGENT-REVIEWERS.md from scratch.

Implements the locked design from bd memory `bfd-design-locked`:
- **6 default specialists** as native Claude Code subagents under `plugins/pr-review-loop/agents/`
- **C+E override semantics** in `discover-agents.sh`: always-on defaults, additive user agents, same-name override, per-default opt-out
- **`# Configuration` section** spec in `AGENT-REVIEWERS.md` (fenced JSON; required `reason` field for `overlap_acknowledged`)
- **Stale-pin detection** at loop start; deterministic version-string compare; exits non-zero unless `--skip-stale-check`
- **Spawning summary** at round 1 start showing default vs user counts and overrides

## What's in this PR (5 commits)

| Commit | What |
|--------|------|
| `893ecf6` | 6 default subagent files (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier) |
| `0851394` | `_load_defaults.sh`, `_parse_configuration.sh`, C+E merge in `discover-agents.sh`; `PR_REVIEW_LOOP_TEST_CHANGED_FILES` env var for offline tests |
| `4843f3c` | SKILL.md: defaults table, C+E semantics, # Configuration spec, stale-pin behavior, spawning summary, pre-loop setup |
| `9225765` | Scripts table updates documenting new helpers |
| `b945eb7` | Test harness — 12 scenarios, 49 assertions, all passing |

Each commit is self-contained and revertible per `pr-review-loop-spike-framing`.

## Test plan

- [x] `tests/test-discover-agents.sh` — 12 scenarios passing (no config / empty config / additive / same-name override / disabled list / version pin match / version pin mismatch / missing-reason rejection / valid-reason / subdirectory config ignored / invalid JSON fallback / combined scenario)
- [x] Verified against this repo's real AGENT-REVIEWERS.md (10 agents spawn: 4 user + 6 defaults)
- [ ] End-to-end loop run on a real PR (this one!)
- [ ] q2h (audit tool / `/pr-review-loop:audit-agents` slash command) — separate ticket; not in this PR

## Acceptance criteria (per bd ticket bfd)

- [x] All 6 defaults exist as native subagents with `name`/`description`/`model`/`color` frontmatter
- [x] C+E semantics implemented in `discover-agents.sh`
- [x] `# Configuration` parser accepts the spec; rejects `overlap_acknowledged` without `reason`
- [x] discover-agents.sh outputs `stale_pin: true|false` (loop enforces exit-non-zero via SKILL.md instructions)
- [x] Stale-pin message references `/pr-review-loop:audit-agents` and `--skip-stale-check`
- [x] Spawning summary documented in SKILL.md
- [x] Merge-readiness summary includes bypass note when `--skip-stale-check` used
- [x] Test harness covers all listed acceptance scenarios

## Dogfooding

This PR is the first dogfood of the pr-review-loop on its own modifications — see bd memory `never-push-to-main-without-explicit-authorization` for the context.

## Refs

- bd ticket: `devon-claude-skills-bfd`
- bd memories: `bfd-design-locked`, `pr-review-loop-spike-framing`, `code-review-philosophy-when-you-touch-code-you`


plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
```diff
@@ -0,0 +1,125 @@
...
+
+AGENTS_DIR="$DEFAULT_AGENTS_DIR"
+if [[ "${1:-}" == "--agents-dir" ]]; then
+    AGENTS_DIR="$(cd "${2:?Usage: _load_defaults.sh [--agents-dir <path>]}" && pwd)"
+fi
+
+if [[ ! -d "$AGENTS_DIR" ]]; then
+    echo "[]"
+    exit 0
+fi
```
devonjones, line 50: 🤖 **Claude Code** (silent-failure-hunter):


Silent failure: missing/unresolvable agents directory swallowed.

```bash
DEFAULT_AGENTS_DIR="$(cd "$SCRIPT_DIR/../../../agents" 2>/dev/null && pwd)" || DEFAULT_AGENTS_DIR=""
...
if [[ ! -d "$AGENTS_DIR" ]]; then
    echo "[]"
    exit 0
fi
```

Two stacked silent fallbacks: the `cd` failure has `2>/dev/null` and the missing-directory branch emits `[]` with exit 0. Together they mean a broken plugin install (agents/ dir missing, moved, or unreadable) produces zero default agents and zero diagnostic output. `discover-agents.sh` consumes this as `DEFAULTS_JSON='[]'` and silently produces a PR review with no specialist reviewers — exit 0, no warning to stderr.

Given this script's whole purpose is to load the 6 shipped defaults, a missing agents/ directory is an installation bug, not a soft-fallback case. Suggest at minimum a stderr warning when the canonical default dir does not resolve and `--agents-dir` was not explicitly passed:

```bash
if [[ ! -d "$AGENTS_DIR" ]]; then
    echo "Warning: default agents directory not found at $DEFAULT_AGENTS_DIR (plugin install may be broken)" >&2
    echo "[]"
    exit 0
fi
```

Or fail loudly when the default path is used and the dir is missing, only emitting `[]` silently when `--agents-dir` was explicitly passed (legitimate empty override).

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
```diff
@@ -0,0 +1,125 @@
...
+    {
+        if (body == "") body = $0
+        else body = body "\n" $0
+    }
+    END {
+        # Skip leading blank lines from body
+        sub(/^\n+/, "", body)
+        printf "{\"name\":\"%s\",\"description\":\"%s\",\"model\":\"%s\",\"color\":\"%s\",\"scope\":\"/\",\"source\":\"%s\",\"instructions\":\"%s\",\"kind\":\"default\"}\n", \
+            json_escape(name), json_escape(description), json_escape(model), json_escape(color), \
+            json_escape(source), json_escape(body)
```
devonjones, line 154: 🤖 **Claude Code** (silent-failure-hunter):


Malformed agent frontmatter is silently passed through as an unnamed agent.

The awk parser initializes `name=""; description=""; model=""; color=""` and unconditionally prints them in END. If an agent .md file has no frontmatter, malformed frontmatter, or is missing the `name:` scalar, the output is `{"name":"","description":"",...}` with no error.

Downstream consequences:
- `discover-agents.sh` will merge a nameless default into the agent list
- The jq override-detection on line 262 (`$user_agent_names | map(. as $n | select($default_agent_names | index($n)))`) will match any user agent named `""` (unlikely) but more importantly the agent will spawn with an empty name — and the pr-review-loop's Task subagent_type mapping will likely fail downstream with a confusing error, far removed from the actual parse failure.

Suggest validating that `name` is non-empty in the awk END block and emitting to stderr + skipping the record:

```awk
END {
    sub(/^\n+/, "", body)
    if (name == "") {
        printf "Warning: %s has no \"name:\" in frontmatter — skipping\n", source > "/dev/stderr"
        exit 0
    }
    printf "{...}\n", ...
}
```

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -1,44 +1,77 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 
+# Plugin version comes from the plugin's own .claude-plugin/plugin.json
+PLUGIN_JSON="$SCRIPT_DIR/../../../.claude-plugin/plugin.json"
+if [[ -f "$PLUGIN_JSON" ]]; then
+    CURRENT_PLUGIN_VERSION="$(jq -r '.version // ""' "$PLUGIN_JSON")"
+else
+    CURRENT_PLUGIN_VERSION=""
+fi
```
devonjones, line 61: 🤖 **Claude Code** (silent-failure-hunter):


Missing or unparseable plugin.json is silently treated as 'no version'.

```bash
PLUGIN_JSON="$SCRIPT_DIR/../../../.claude-plugin/plugin.json"
if [[ -f "$PLUGIN_JSON" ]]; then
    CURRENT_PLUGIN_VERSION="$(jq -r '.version // ""' "$PLUGIN_JSON")"
else
    CURRENT_PLUGIN_VERSION=""
fi
```

If `plugin.json` is missing (broken install) or lacks `.version` (config bug), `CURRENT_PLUGIN_VERSION` becomes `""` with no log. This then flows into the stale-pin check on line 295:

```jq
(($config.defaults_version_checked // null) != $current_version) as $stale_pin
```

When current_version is `""`, any non-empty `defaults_version_checked` triggers `stale_pin: true` (likely misinterpreted by the user as 'I need to bump my pin', not 'the plugin install is broken'). And the output `current_plugin_version: ""` is the only signal — which Test 1 on line 147 asserts is "present" but allows empty-string via its check (`!= null and != ""`), so a regression that breaks plugin.json detection would also break the test, but only obliquely.

Suggest at least a stderr warning when plugin.json is missing or lacks `.version`:

```bash
if [[ -f "$PLUGIN_JSON" ]]; then
    CURRENT_PLUGIN_VERSION="$(jq -r '.version // ""' "$PLUGIN_JSON")"
    if [[ -z "$CURRENT_PLUGIN_VERSION" ]]; then
        echo "Warning: plugin.json at $PLUGIN_JSON has no .version field" >&2
    fi
else
    echo "Warning: plugin.json not found at $PLUGIN_JSON — stale_pin checks will be unreliable" >&2
    CURRENT_PLUGIN_VERSION=""
fi
```

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -153,47 +184,121 @@ parse_agent_file() {
...
-# Merge agents: lower scope (more specific path) wins on name collision.
-# Sort by scope depth (deeper first), then deduplicate by name keeping first.
-# Collect context sections (no dedup needed — all are passed through).
-# Build a JSON array of changed files for scoping
+# Build JSON array of changed files for scoping
 CHANGED_FILES_JSON=$(jq -n --arg files "$CHANGED_FILES" '$files | split("\n") | map(select(. != ""))')
 
-printf '%s\n' "$ALL_SECTIONS" | jq -s --argjson files "$CHANGED_FILES_JSON" '
+# Load defaults from plugins/pr-review-loop/agents/*.md
+DEFAULTS_JSON=$("$SCRIPT_DIR/_load_defaults.sh")
```
devonjones, line 209: 🤖 **Claude Code** (silent-failure-hunter):


Silent zero-default-agents state is not surfaced.

```bash
DEFAULTS_JSON=$("$SCRIPT_DIR/_load_defaults.sh")
```

`_load_defaults.sh` can legitimately return `[]` with exit 0 in two cases (missing agents dir, empty agents dir). When that happens here, `DEFAULTS_JSON='[]'` flows through the jq merge logic and produces a final output with `default_count: 0`, `spawned_count: 0` (assuming no user agents) — exit 0, no warning. The pr-review-loop will then spawn zero reviewers and proceed.

For a tool whose primary contract per the PR ("6 default reviewer agents") is loading defaults, getting zero defaults silently is the worst kind of silent failure: the user gets a 'successful' review pass with no actual review. Combined with the upstream fallbacks in `_load_defaults.sh` (see other comments), this is the spot where the silence becomes user-visible.

Suggest a sanity check immediately after the load:

```bash
DEFAULTS_JSON=$("$SCRIPT_DIR/_load_defaults.sh")
DEFAULTS_COUNT=$(echo "$DEFAULTS_JSON" | jq 'length')
if [[ "$DEFAULTS_COUNT" -eq 0 ]]; then
    echo "Warning: no default reviewer agents loaded from $SCRIPT_DIR/../../../agents — plugin install may be broken or all defaults disabled." >&2
fi
```

This is non-fatal (legitimate cases exist — e.g., all-disabled via config), but it makes the silent zero state visible.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -0,0 +1,84 @@
...
+    exit 0
+fi
+
+# Validate JSON shape via jq. If parse fails, surface error + return {} so
+# the loop can proceed (but emit a warning to stderr).
+if ! echo "$RAW_JSON" | jq -e . > /dev/null 2>&1; then
+    echo "Warning: # Configuration section in $FILE contains invalid JSON" >&2
+    echo "{}"
+    exit 0
+fi
```
devonjones, line 70: 🤖 **Claude Code** (silent-failure-hunter):


Invalid # Configuration JSON degrades silently to 'all defaults, stale_pin=true' with only a stderr warning.

```bash
if ! echo "$RAW_JSON" | jq -e . > /dev/null 2>&1; then
    echo "Warning: # Configuration section in $FILE contains invalid JSON" >&2
    echo "{}"
    exit 0
fi
```

The warning is good, but the failure mode is high-impact: if a user's `disabled` list has a JSON typo, every reviewer they intended to disable will still spawn — silently. The user sees the reviewers run, doesn't notice the buried stderr line in a long-running pr-review-loop, and concludes their config 'works.'

Two specific concerns:
1. The warning omits *what* was invalid — no jq error message is captured. Suggest piping jq's stderr to a tempfile and including the first line of its error in the warning:
   ```bash
   JQ_ERR=$(echo "$RAW_JSON" | jq -e . 2>&1 >/dev/null) || {
       echo "Warning: # Configuration section in $FILE contains invalid JSON: ${JQ_ERR%%$'\n'*}" >&2
       echo "{}"
       exit 0
   }
   ```
2. Test 11 (line 326 of test-discover-agents.sh) only asserts `defaults still spawn after invalid config` and `stale_pin true` — it does *not* assert that a warning was emitted to stderr. The fallback's silence is not test-locked, so a future refactor could remove the warning without test failure. Recommend adding `assert_stderr_contains 'invalid JSON warning' "$t11_err" 'invalid JSON'` to that test.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -153,47 +184,121 @@ parse_agent_file() {
...
+    # Defaults that get replaced because user has same-name agent
+    ($user_agent_names | map(. as $n | select($default_agent_names | index($n)))) as $overridden_default_names |
+
+    # Disabled defaults (from configuration)
+    ($config.disabled // []) as $disabled_defaults |
+
+    # ---- 4. Effective defaults: not overridden, not disabled ----
+    ($defaults
+        | map(select(.name as $n | ($overridden_default_names | index($n)) == null))
+        | map(select(.name as $n | ($disabled_defaults | index($n)) == null))
```
devonjones, line 302: 🤖 **Claude Code** (type-design-analyzer):


type: MEDIUM — `disabled_defaults` invariant not enforced against the default set.

The `disabled` list in `# Configuration` is filtered through at line 270 (`map(select(.name as $n | ($disabled_defaults | index($n)) == null))`) but is never cross-checked against `($defaults | map(.name))`. A user who writes `"disabled": ["pr-test-analyser"]` (British spelling typo) silently disables nothing — no warning, no error, and the agent runs as if unconfigured. The output `configuration.disabled_defaults` echoes the typo back, so a downstream auditor can't distinguish typos from valid skips.

Invariants identified:
- Each entry in `disabled` is a name of a shipped default agent.
- Empty / unknown names should be diagnosed, not silently dropped.

Ratings: Encap 6/10 · Invar-Expr 4/10 · Invar-Useful 8/10 · Invar-Enforce 3/10

Suggested minimal improvement: after the merge step, compute `($disabled_defaults - $default_agent_names)` and emit a stderr warning listing any unknown names. Same pattern as the subdirectory-config warning already in place at lines 219-224. Or surface an `unknown_disabled` field in the output `configuration` block so the audit tool (q2h) can prompt about it.

Parallels the existing strong `overlap_acknowledged.reason` check (in `_parse_configuration.sh`) — that one rejects bad input loudly; `disabled` accepts bad input silently. The asymmetric enforcement is the smell.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
```diff
@@ -0,0 +1,125 @@
...
+    }
+    # After frontmatter closes: accumulate body
+    {
+        if (body == "") body = $0
+        else body = body "\n" $0
+    }
+    END {
+        # Skip leading blank lines from body
+        sub(/^\n+/, "", body)
+        printf "{\"name\":\"%s\",\"description\":\"%s\",\"model\":\"%s\",\"color\":\"%s\",\"scope\":\"/\",\"source\":\"%s\",\"instructions\":\"%s\",\"kind\":\"default\"}\n", \
```
devonjones, line 152: 🤖 **Claude Code** (pr-test-analyzer):


**Coverage gap (criticality 8): malformed/missing frontmatter produces a silent zero-value default.**

The awk parser here unconditionally emits a JSON object even when frontmatter is malformed. Two reproducible failure modes that no test exercises:

1. **No closing \`---\`** (frontmatter never closes): emits agent with empty \`instructions\` (body is never accumulated because \`in_frontmatter\` stays 1). Verified locally: returns \`{"name":"broken","description":"...","model":"opus","color":"","instructions":""}\`. A default with empty instructions would silently spawn and produce useless reviews.

2. **No frontmatter at all** (plain markdown file dropped in \`agents/\`): emits agent with \`name=""\`. Verified locally. An agent with empty \`name\` (a) participates in override/disabled matching as \`""\`, (b) could collide with another empty-named file, (c) breaks downstream consumers that key on name.

Suggested tests in \`test-discover-agents.sh\` (invoke \`_load_defaults.sh --agents-dir <tmpdir>\` directly):

- Test 13: agents dir with one file missing closing \`---\` → either rejected with non-zero exit, or assert the documented fallback shape. Currently neither is guaranteed.
- Test 14: agents dir with a file missing required \`name\` field → expect rejection or skip, not silent emission of \`name=""\`.

Both are real regressions waiting to happen the next time someone hand-edits an agent file. The fix is straightforward (validate required fields; reject if frontmatter unterminated) but the tests should be in place first to lock the contract.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -0,0 +1,84 @@
...
+# Extract the fenced JSON block within the # Configuration H1 section.
+# Logic:
+#   - On `# Configuration`, enter configuration section
+#   - On any other `# `, leave configuration section
+#   - Inside configuration section, when we see ```json, capture lines
+#     until the closing ```
+RAW_JSON="$(awk '
+    BEGIN { in_config = 0; in_fence = 0 }
+    /^# Configuration[[:space:]]*$/ { in_config = 1; next }
+    /^# / && !/^# Configuration[[:space:]]*$/ { in_config = 0 }
```
devonjones: 🤖 **Claude Code** (code-simplifier):


Redundant guard. Line 49 already matches `/^# Configuration[[:space:]]*$/` and runs `next`, so line 50 never sees the Configuration heading. The `!/^# Configuration[[:space:]]*$/` filter is dead — `/^# /` alone is equivalent and easier to read. Verified behavior-identical on a sample doc.

```suggestion
    /^# / { in_config = 0 }
```

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -0,0 +1,84 @@
...
+# bd memory bfd-design-locked: reason is REQUIRED.
+INVALID_ENTRIES="$(echo "$RAW_JSON" | jq -r '
+    .overlap_acknowledged // {}
+    | to_entries
+    | map(select(.value.reason == null or .value.reason == ""))
+    | map(.key)
+    | join(", ")
+')"
+
+if [[ -n "$INVALID_ENTRIES" ]]; then
```
devonjones, line 95: 🤖 **Claude Code** (pr-test-analyzer):


**Coverage gap (criticality 7): \`reason: ""\` and \`reason: null\` are not exercised by any test.**

Test 8 covers the case where the \`reason\` key is *missing* from an \`overlap_acknowledged\` entry. But the jq validator on line 73-76 also rejects \`reason == null\` and \`reason == ""\`. Neither of those explicit-but-empty forms is tested.

Why this matters: an author writing \`"reason": ""\` (or \`"reason": null\`) is a much more likely real-world mistake than omitting the key entirely — JSON editors and templating often leave empty strings behind. The current implementation handles them correctly (I verified \`reason: ""\` triggers exit 1), but if someone refactors the jq selector to e.g. \`.value.reason | not\` or drops the empty-string check, no test would catch the regression.

Suggested test (criticality 7):
- Test 8b: \`overlap_acknowledged.x.reason == ""\` → expect exit 1, stderr contains "missing required 'reason'"
- Test 8c: \`overlap_acknowledged.x.reason == null\` → expect exit 1

These take ~10 lines each as variations on Test 8.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -153,47 +184,121 @@ parse_agent_file() {
...
+
+    # ---- 5. User agents: tagged as override if name matches a default, else plain user ----
+    ($user_agents_raw
+        | map(.name as $n | . + {kind: (if ($default_agent_names | index($n)) != null then "user-override" else "user" end)})
+        | map(.scope as $s | . + {changed_files: (
+            if $s == "/" then $files
+            else [$files[] | select(startswith($s | ltrimstr("/")))]
+            end
+          )})
+        | map({name, scope, source, instructions, kind, changed_files})
```
devonjones, line 327: 🤖 **Claude Code** (type-design-analyzer):


type: MEDIUM — asymmetric agent record shape between defaults and user agents.

Default agents (from `_load_defaults.sh`) carry `{name, description, model, color, scope, source, instructions, kind}`. User agents at line 288 are explicitly projected down to `{name, scope, source, instructions, kind, changed_files}` — `description`, `model`, and `color` are dropped.

The SKILL.md documentation of the JSON shape (lines 18-35 of this file) declares: `agents: [{name, scope, source, instructions, model?, color?, kind, changed_files}, ...]` — `model` and `color` are documented as optional, which is consistent with the user/default asymmetry. So far so good.

But: `description` is silently dropped for user agents AND is not mentioned in the doc-comment JSON shape at all. The audit subcommand (q2h, blocked on this work) is the obvious consumer that needs descriptions to render a 'these are the agents in your repo' summary. A user agent that's documented in AGENT-REVIEWERS.md's body prose still has no `description` field on its JSON record, which forces the audit tool to either (a) re-parse the markdown or (b) display defaults with descriptions and user agents without.

Invariants identified:
- Every agent record has the same field set, regardless of source.
- Optional fields are documented as optional in the shape comment.

Ratings: Encap 7/10 · Invar-Expr 5/10 · Invar-Useful 6/10 · Invar-Enforce 6/10

Suggested minimal improvement: either (a) extend the user-agent parser to extract a leading paragraph or first sentence under the H2 as `description`, or (b) explicitly note in the shape comment at line 18 that `description` is default-only. Picking one removes the implicit shape asymmetry. Probably (b) is the smaller move for this PR; (a) is the better long-term answer and might warrant a beads follow-up if Devon agrees.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -153,47 +184,121 @@ parse_agent_file() {
...
+
+    # ---- 3. Determine overrides ----
+    # Defaults that get replaced because user has same-name agent
+    ($user_agent_names | map(. as $n | select($default_agent_names | index($n)))) as $overridden_default_names |
+
+    # Disabled defaults (from configuration)
+    ($config.disabled // []) as $disabled_defaults |
+
+    # ---- 4. Effective defaults: not overridden, not disabled ----
+    ($defaults
```
devonjones, line 300: 🤖 **Claude Code** (pr-test-analyzer):


**Coverage gap (criticality 7): a default that is BOTH \`disabled\` AND has a same-name user override is not tested.**

The jq pipeline on lines 268-278 filters defaults by removing the union of (overridden ∪ disabled). The user agent with the same name still spawns as \`user-override\`. Verified locally with \`disabled: ["code-reviewer"]\` + a user \`## code-reviewer\` block: result is \`code-reviewer kind=user-override\` plus 5 other defaults (6 total).

That's a plausible reading of the spec, but it's *ambiguous user intent*: did they want to disable code-reviewer entirely, or replace it? The current behavior silently picks "replace", which contradicts the disabled list. Either:

- **(a)** Treat \`disabled\` as authoritative and skip the user override too (in which case the test should assert 5 agents, no \`code-reviewer\`).
- **(b)** Treat \`override\` as authoritative (current behavior) and the test should lock that in.
- **(c)** Reject as a configuration error.

Whichever you pick, this PR ships the merge logic and a test harness — the design decision should be encoded in a test scenario so a future refactor can't quietly flip it. The prompt for this reviewer explicitly called this case out, so it's not a hypothetical.

Suggested Test 13: disabled + same-name override → assert chosen behavior (kind, count, presence/absence in \`agents\`).

plugins/pr-review-loop/skills/pr-review-loop/tests/test-discover-agents.sh
```diff
@@ -0,0 +1,403 @@
...
+    local td
+    td="$(mktemp -d)"
+    (
+        cd "$td"
+        git init -q
+        git config user.email "test@example.com"
+        git config user.name "Test"
+    )
+    echo "$td"
+}
```
devonjones, line 63: 🤖 **Claude Code** (shell-script-reviewer):


Test harness does not guarantee temp dir cleanup on failure. Each test ends with `rm -rf "$repo"` (e.g., line 148, 166, 184...) but with `set -euo pipefail` (line 15), any unexpected failure between `make_temp_repo` and the trailing `rm -rf` leaves behind `/tmp/tmp.XXX` directories (and the `stdout_file`/`stderr_file` mktemps in `run_discover`, lines 62-63). Note that `assert_jq`/`assert_exit`/`assert_stderr_contains` swallow failures via counters so they won't trip `set -e` — but any other unexpected error (heredoc failure, jq segfault, mktemp -d returning non-zero on a full /tmp, an interrupted CI run via Ctrl-C, etc.) will leak. Consider an EXIT trap that tracks created temp dirs in an array and `rm -rf`s them on exit. Minor: not a blocker, but the task brief explicitly called this out as a concern to verify.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones, line 70: 🤖 **Claude Code** (type-design-analyzer):


type: MEDIUM — asymmetric schema enforcement on the `# Configuration` object.

The parser strongly enforces one invariant (`overlap_acknowledged.<name>.reason` required, line 79-82) but silently accepts unknown top-level keys. A user who writes `"diabled": ["foo"]` (missing 'd') gets no warning — the field is dropped on the floor and `disabled_defaults` defaults to `[]` downstream.

The valid top-level keys are exactly three: `defaults_version_checked`, `disabled`, `overlap_acknowledged`. They're documented in SKILL.md (lines 627-649) and in the file header comment here (lines 11-19), but the schema is enforced nowhere — it lives only in prose.

Invariants identified:
- Top-level keys come from a small, fixed allowlist.
- Unknown keys at the top level are user errors that should surface.

Ratings: Encap 7/10 · Invar-Expr 4/10 · Invar-Useful 7/10 · Invar-Enforce 5/10

Suggested minimal improvement: after the JSON-validity check at line 63 and before exit, compute the set difference between the parsed object's keys and the allowed set, and emit a stderr warning if non-empty:

```bash
UNKNOWN="$(echo "$RAW_JSON" | jq -r '
    ["defaults_version_checked", "disabled", "overlap_acknowledged"] as $known
    | (keys - $known) | join(", ")
')"
if [[ -n "$UNKNOWN" ]]; then
    echo "Warning: unknown # Configuration keys in $FILE: $UNKNOWN" >&2
fi
```

Keep the exit code 0 so a typo doesn't break the loop, just surface the warning. Same vibe as the subdirectory-config warning in `discover-agents.sh`. This closes the schema enforcement asymmetry — the parser already validates one nested invariant strictly; extending the same care to the top-level shape is consistent.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones, line 209: 🤖 **Claude Code** (pr-test-analyzer):


**Coverage gap (criticality 6): hierarchical scope of user agents in subdirectory AGENT-REVIEWERS.md files is not exercised together with defaults.**

Test 10 covers a subdirectory \`# Configuration\` being ignored. But there's no test that exercises a subdirectory \`# Agents\` block providing a user agent scoped to that subdirectory while defaults still merge in.

I verified the current behavior works locally: a \`plugins/foo/AGENT-REVIEWERS.md\` with \`## scoped-agent\` plus a root \`AGENT-REVIEWERS.md\` with \`## sub-agent\` produces both as \`kind=user\` with scopes \`/plugins/foo/\` and \`/\`, and the 6 defaults still merge in.

This matters because the *merge with defaults* on line 200 (\`DEFAULTS_JSON=$("$SCRIPT_DIR/_load_defaults.sh")\`) and subsequent jq merge logic is new in this PR. A regression that breaks the interaction between subdir-scoped user agents and the new default-merge wouldn't be caught.

Suggested Test 13/14: nested AGENT-REVIEWERS.md with user agents at a subdir scope → assert (a) scope strings, (b) 6 defaults still spawn, (c) the subdir agent's \`changed_files\` filtering works (a file outside the subdir should give that agent zero \`changed_files\`).

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
```diff
@@ -0,0 +1,125 @@
...
+            seen_frontmatter = 1
+            next
+        } else if (in_frontmatter) {
+            in_frontmatter = 0
+            next
+        }
+    }
+    # Inside frontmatter: parse top-level scalars only
+    in_frontmatter {
+        if (match($0, /^([a-zA-Z_][a-zA-Z0-9_]*):[[:space:]]*(.*)$/, m)) {
```
devonjones: 🤖 **Claude Code** (code-reviewer):


**bug:** Portability — `match(/usr/bin/zsh, /regex/, m)` with a capture-group array (3rd arg) is a **GNU awk extension**. BSD awk (the default `/usr/bin/awk` on macOS) only accepts `match(string, regex)` and does not populate a capture-group array.

**Concrete impact:** On macOS, this awk block silently no-ops (the `match` call still returns the position, so the `if` body runs, but `m[1]` and `m[2]` are empty). Result: every frontmatter scalar (`name`, `description`, `model`, `color`) is the empty string, and `_load_defaults.sh` emits 6 nameless default agents. `discover-agents.sh` then sees `name == ""` for every default — overrides, disabled-list lookups, and same-name detection all break.

The SKILL.md Prerequisites section explicitly supports macOS (`brew install jq`) but does not document a `gawk` requirement, so end users on macOS will hit this.

**Suggested fix (drop the capture-group form, use `substr` + a second match like `discover-agents.sh` already does):**

```awk
if (match($0, /^[a-zA-Z_][a-zA-Z0-9_]*:[[:space:]]*/)) {
    key_end = RSTART + RLENGTH - 1
    # find the colon to split key/value
    colon = index($0, ":")
    key = substr($0, 1, colon - 1)
    val = substr($0, key_end + 1)
    # ... (existing quote-stripping unchanged)
}
```

Alternative: document gawk as a Prerequisite in SKILL.md (`brew install gawk` on macOS) and update the shebang or call `gawk` explicitly. The substr fix is simpler and keeps the script portable.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -573,9 +589,97 @@ When using Claude fallback:
...
+  }
+}
+```
+````
+
+Fields:
+
+- **`defaults_version_checked`** — plugin version (matches the value in `.claude-plugin/plugin.json`) whose defaults the user has reviewed. The `/pr-review-loop:audit-agents` tool (q2h) bumps this when the user accepts/rejects each recommendation. Same version namespace as the plugin.
+- **`disabled`** — list of default agent names the user does NOT want spawned. Per-name opt-out.
+- **`overlap_acknowledged`** — map from a user agent name to `{ overlaps_with, reason }`. Both agents continue to spawn; this entry just documents intentional duplication so the audit tool doesn't recommend renaming. **`reason` is REQUIRED** — the parser rejects entries without it. Purpose: future readers of `AGENT-REVIEWERS.md` see why both agents are intentionally running, not perceive it as cleanup-eligible cruft.
```


clarity: this bullet says the same thing twice. The middle clause already explains the purpose: "this entry just documents intentional duplication so the audit tool doesn't recommend renaming." The trailing "Purpose: future readers ... not perceive it as cleanup-eligible cruft" restates it from a reader's POV. Drop the Purpose sentence.

Verbose: `...this entry just documents intentional duplication so the audit tool doesn't recommend renaming. **`reason` is REQUIRED** — the parser rejects entries without it. Purpose: future readers of `AGENT-REVIEWERS.md` see why both agents are intentionally running, not perceive it as cleanup-eligible cruft.`

Terse: `...this entry documents intentional duplication so the audit tool doesn't recommend renaming. **`reason` is REQUIRED** — the parser rejects entries without it.`

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -573,9 +589,97 @@ When using Claude fallback:
...
+      "reason": "PCI-compliance-specific scope; we want both running because the default is general security"
+    }
+  }
+}
+```
+````
+
+Fields:
+
+- **`defaults_version_checked`** — plugin version (matches the value in `.claude-plugin/plugin.json`) whose defaults the user has reviewed. The `/pr-review-loop:audit-agents` tool (q2h) bumps this when the user accepts/rejects each recommendation. Same version namespace as the plugin.
```


clarity: "Same version namespace as the plugin." is redundant with the preceding clause that already says the value matches `.claude-plugin/plugin.json`. Drop the tail.

Verbose: `plugin version (matches the value in `.claude-plugin/plugin.json`) whose defaults the user has reviewed. The `/pr-review-loop:audit-agents` tool (q2h) bumps this when the user accepts/rejects each recommendation. Same version namespace as the plugin.`

Terse: `plugin version (matches `.claude-plugin/plugin.json`) whose defaults the user has reviewed. The `/pr-review-loop:audit-agents` tool (q2h) bumps this when the user accepts/rejects each recommendation.`

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -252,6 +253,21 @@ The review loop should **not** override branch protections or bypass repo-define
 4. **ALWAYS use `--wait` flag** when checking for comments - this ensures proper 5-minute polling
 5. **PR creation automatically triggers Gemini review** - use `get-review-comments.sh --wait` to wait for the first review
 
+### Pre-Loop Setup (do once, before round 1)
+
+Before any review round, perform setup:
```


clarity: the heading already says "Pre-Loop Setup (do once, before round 1)". The "Before any review round, perform setup:" line restates the heading. Drop it and let the numbered list speak for itself.

Verbose: `### Pre-Loop Setup (do once, before round 1)\n\nBefore any review round, perform setup:\n\n1. **Discover and merge agents.** ...`

Terse: `### Pre-Loop Setup (do once, before round 1)\n\n1. **Discover and merge agents.** ...`

plugins/pr-review-loop/agents/silent-failure-hunter.md
```diff
@@ -0,0 +1,73 @@
...
+description: Reviewer that hunts silent failures, inadequate error handling, and inappropriate fallback behavior in PR diffs. Use when reviewing code that adds or modifies try-catch blocks, error handlers, fallback logic, optional chaining, or any path where errors could be suppressed without surfacing.
+model: opus
+color: yellow
+---
+
+You are a reviewer hunting silent failures and inadequate error handling in PR diffs.
+
+**Your focus:** Error handling code in files touched by this PR — try/catch (or try/except, Result types, error callbacks, conditional error branches, fallback logic, optional chaining, null coalescing). Find places where errors are swallowed, masked by fallbacks, or surfaced too vaguely to be actionable.
+
+**Scope rule (touch-it-you-own-it):** Issues in files modified by this PR are in scope, regardless of whether the error-handling defect pre-existed or was introduced by this PR. Issues in files NOT in the diff are out of scope — they belong to a different PR.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


clarity: the trailing "— they belong to a different PR" is redundant with "out of scope". The other five agent files state this scope rule without the tail (e.g., code-reviewer.md line 17, comment-analyzer.md line 12). Drop for consistency and brevity.

Verbose: `Issues in files NOT in the diff are out of scope — they belong to a different PR.`

Terse: `Issues in files NOT in the diff are out of scope.`

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
devonjones: Fixed - added stderr warnings when agents dir missing or empty (no more silent zero-defaults install failure)

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
devonjones: Fixed - END-block validation now rejects empty/missing name with a stderr warning, and the file is skipped (no nameless agent reaches downstream)

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - warns to stderr when plugin.json is missing OR when its version field is empty

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - emits a stderr warning when _load_defaults.sh returns zero defaults (broken plugin install no longer silent)

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - captures jq's actual parse error (line/column) in the warning. Test 11 still verifies the graceful-fallback path; the warning content is now informative

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - warns on disabled names that don't match any default (asymmetric strictness reconciled with the reason-required check). Test 16 locks in the behavior

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
devonjones: Fixed - _load_defaults.sh now skips files with malformed/missing frontmatter (no name) with a stderr warning. The validation is exercised indirectly via the existing 12 scenarios (all 6 shipped defaults pass through the new validator)

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - simplified to just /^# / { in_config = 0 } (behavior-identical; the Configuration heading is matched by the earlier rule with next, so this catches every other H1)

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - Test 14 added covering both empty-string and explicit-null reason; jq filter now uses (.value.reason // "") | length == 0 so all three (missing key, null, empty string) are uniformly rejected

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - added inline comment at the user_agents block documenting the shape asymmetry: defaults carry description/model/color from frontmatter; user agents carry only what AGENT-REVIEWERS.md provides. SKILL.md spawn template handles missing model by falling back to orchestrator default

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - added inline comment at the effective_defaults block documenting precedence (user-override wins; default filtered by EITHER rule). Test 13 locks the behavior in

plugins/pr-review-loop/skills/pr-review-loop/tests/test-discover-agents.sh
devonjones: Fixed - added trap-based cleanup using TEMP_PATHS array. Temp dirs and the mktemp stdout/stderr files are now tracked and cleaned on EXIT (set -e abort, Ctrl-C, etc.)

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - warn on unknown top-level keys with the allowed-key list. Test 15 covers the diabled-typo case

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Out of scope - tracked in devon-claude-skills-tdj. The hierarchical-subdir x default-merge interaction needs a few new test fixtures and surfaces a design question (does a subdir override of a default apply repo-wide or per-scope?). Filed as P3 for a focused follow-up PR

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
devonjones: Fixed - rewrote the frontmatter parser to use portable substr-based parsing (no gawk capture-group syntax). BSD awk on macOS now produces identical output. Verified with the existing test harness (49 -> 65 assertions, all passing)

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - merged the Purpose rationale into the same sentence as the required-reason explanation; no duplication

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - dropped the redundant 'Same version namespace as the plugin' tail (the preceding parenthetical already states that)

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - dropped 'Before any review round, perform setup:' line; the heading carries the intent

plugins/pr-review-loop/agents/silent-failure-hunter.md
devonjones: Fixed - dropped the trailing '— they belong to a different PR' for consistency with the other 5 default agents (which state 'out of scope' alone)

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
```diff
@@ -0,0 +1,166 @@
...
+        if (!seen_frontmatter) {
+            print "Warning: " source " has no YAML frontmatter (no --- markers); skipping" > warn_target
+            exit 0
+        }
+        if (in_frontmatter) {
+            print "Warning: " source " frontmatter never closes (missing second ---); skipping" > warn_target
+            exit 0
+        }
+        if (name == "") {
+            print "Warning: " source " has empty or missing `name` field in frontmatter; skipping" > warn_target
```
devonjones, line 149: 🤖 **Claude Code** (pr-test-analyzer):


**Coverage gap (criticality 7): the new frontmatter validators in `_load_defaults.sh` (lines 142-153) have zero direct test coverage.**

Round 1's reply stated the validation is "exercised indirectly via the existing 12 scenarios (all 6 shipped defaults pass through the new validator)". That's only true for the *success* path. The three new skip-with-warning paths added in this commit are never reached by any test:

- Line 143: `has no YAML frontmatter (no --- markers); skipping` — no test feeds a bare-markdown file into `_load_defaults.sh`.
- Line 147: `frontmatter never closes (missing second ---); skipping` — no test feeds a half-frontmatter file.
- Line 151: `empty or missing `name` field in frontmatter; skipping` — no test feeds a frontmatter-without-name file.

Verified: `tests/test-discover-agents.sh` is the only test file, and it exercises `discover-agents.sh` (which always loads the real, valid 6 defaults via `_load_defaults.sh` with no `--agents-dir`). None of the 16 scenarios construct a malformed agent file.

Why this matters: these validators are precisely what round-1 asked for, and they're now the *only* thing standing between a fat-fingered hand-edit of an agent file and the silent-failure regression I originally flagged. If a future refactor drops the `exit 0` after the warning and lets parsing fall through (or swaps the `name == ""` check for a typo'd `namee`), every existing test would still pass.

Concrete fix (~30 lines, no new infra needed): a tiny `tests/test-load-defaults.sh` that builds a temp agents dir with three crafted files, invokes `_load_defaults.sh --agents-dir <tmpdir>`, asserts (a) jq output excludes the bad files, (b) stderr contains each expected warning string. The `--agents-dir` flag is already plumbed for exactly this purpose.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -1,44 +1,84 @@
...
 # {
-#   "agents": [
-#     {"name": "security-reviewer", "scope": "/", "source": "/AGENT-REVIEWERS.md", "instructions": "..."},
-#     {"name": "sql-injection-reviewer", "scope": "/services/api/", "source": "/services/api/AGENT-REVIEWERS.md", "instructions": "..."}
-#   ],
-#   "context": [
-#     {"scope": "/", "section": "Guidelines", "content": "..."},
-#     {"scope": "/services/api/", "section": "Guidelines", "content": "..."}
-#   ]
+#   "agents": [{name, scope, source, instructions, model?, color?, kind, changed_files}, ...],
```
devonjones: 🤖 **Claude Code** (type-design-analyzer):


type: MEDIUM — the documented output shape at line 20 omits 'description', which the inline comment at lines 318-324 then acknowledges defaults DO carry.

The shape comment at line 20 declares:
```
agents: [{name, scope, source, instructions, model?, color?, kind, changed_files}, ...]
```

But the inline asymmetry note at lines 318-324 (added by this commit) explicitly says defaults additionally carry description, model, color from frontmatter. Round 1's reply also confirmed defaults carry description but user agents don't. So 'description' is a real default-only field — yet the API contract at line 20 makes it look like it doesn't exist on any record.

A consumer reading line 20 to know what fields are queryable would never know defaults have a description. The spawn template in SKILL.md doesn't use description, so today this is harmless — but the audit subcommand (q2h, blocked on this work) was the explicit reason raised in round 1: it needs descriptions to render a 'these are the agents in your repo' summary. It will hit `.description` and either get null (if jq is lenient) or be confused that some records have it and some don't, with no documentation telling it which.

Invariants identified:
- The output shape comment is a contract; every field that any record carries should be listed (with optionality marked).
- 'description' is default-only and should be marked 'description?' (or '(defaults only)') in the shape comment, mirroring the existing 'model?, color?' treatment.

Ratings: Encap 7/10 · Invar-Expr 4/10 · Invar-Useful 7/10 · Invar-Enforce 5/10

Suggested minimal improvement: add 'description?' to the field list at line 20 alongside 'model?, color?'. The inline note at 318-324 is already honest about the source of the asymmetry; extending the shape comment to list the field closes the loop. Two-character fix to the docstring.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -0,0 +1,104 @@
...
+fi
+
+# Validate JSON shape via jq. If parse fails, capture the actual jq error so
+# the user can see WHERE the malformed JSON is, then fall back to {} so the
+# loop can proceed. Disable `set -e` around the parse so a failure surfaces
+# the diagnostic instead of aborting the script.
+set +e
+JQ_PARSE_ERR="$(echo "$RAW_JSON" | jq -e . 2>&1 > /dev/null)"
+JQ_PARSE_EXIT=$?
+set -e
```
devonjones: 🤖 **Claude Code** (code-simplifier):


The `set +e` / capture-`$?` / `set -e` dance is unnecessary here. Inside an \`if\` condition (or after \`!\`), command failure doesn't trigger \`set -e\` abort, so you can capture stderr and exit status in one shot:

\`\`\`suggestion
if ! JQ_PARSE_ERR=\"\$(echo \"\$RAW_JSON\" | jq -e . 2>&1 > /dev/null)\"; then
    echo \"Warning: # Configuration section in \$FILE contains invalid JSON: \$JQ_PARSE_ERR\" >&2
    echo \"{}\"
    exit 0
fi
\`\`\`

Behavior-identical: assignment from \`\$()\` inherits the command's exit status, the \`if !\` makes set-e tolerate the failure path, and the success path produces a no-op echo (jq -e prints nothing on stderr when valid). Drops two set-state mutations and one intermediate variable.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
```diff
@@ -0,0 +1,166 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# Normalize the default path so the `..` segments don't leak into output.
+# Warn if the agents directory can't be resolved — a broken install would
+# otherwise silently load zero defaults.
+if DEFAULT_AGENTS_DIR="$(cd "$SCRIPT_DIR/../../../agents" 2>/dev/null && pwd)"; then
+    :
+else
+    echo "Warning: default agents dir not found at $SCRIPT_DIR/../../../agents — plugin install may be incomplete" >&2
+    DEFAULT_AGENTS_DIR=""
+fi
```
devonjones, line 39: 🤖 **Claude Code** (code-simplifier):


The empty \`then : else ...\` branch reads as a forgotten TODO. Inverting the condition with \`!\` collapses it to the natural form:

\`\`\`suggestion
if ! DEFAULT_AGENTS_DIR=\"\$(cd \"\$SCRIPT_DIR/../../../agents\" 2>/dev/null && pwd)\"; then
    echo \"Warning: default agents dir not found at \$SCRIPT_DIR/../../../agents — plugin install may be incomplete\" >&2
    DEFAULT_AGENTS_DIR=\"\"
fi
\`\`\`

Behavior-identical (assignment from \`\$()\` propagates the command's exit status; \`if !\` makes set-e tolerate the failure path). Drops the no-op \`:\` arm.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -202,6 +351,18 @@ printf '%s\n' "$ALL_SECTIONS" | jq -s --argjson files "$CHANGED_FILES_JSON" '
                 end
             ) | .result |
             map({section, scope, source, content})
-        )
+        ),
+        configuration: {
+            defaults_version_checked: ($config.defaults_version_checked // null),
+            current_plugin_version: $current_version,
+            stale_pin: $stale_pin,
+            disabled_defaults: $disabled_defaults,
```
devonjones, line 354: 🤖 **Claude Code** (type-design-analyzer):


type: MEDIUM — `configuration.disabled_defaults` echoes the raw user list verbatim, including names that didn't actually disable anything.

This commit added a stderr warning at lines 245-256 when the disabled list contains names that don't match any default — good. But the warning is the ONLY surface for this. Line 359 still emits `disabled_defaults: $disabled_defaults` (the raw user input), so given:

```json
{ "disabled": ["nonexistent-reviewer", "pr-test-analyzer"] }
```

The output is:
```json
{ "disabled_defaults": ["nonexistent-reviewer", "pr-test-analyzer"] }
```

…even though only `pr-test-analyzer` was actually filtered out of the spawned agents. A downstream consumer (the audit subcommand q2h, or any tool reading the JSON to render 'what's disabled in this repo') would believe `nonexistent-reviewer` is one of the disabled defaults. The stderr warning is invisible to JSON consumers — only the human watching the run sees it.

Invariants identified:
- `configuration.disabled_defaults` should mean 'names of defaults actually filtered out', not 'names the user wrote in the disabled list'.
- The intersection of `disabled_defaults` and `overridden_default_names` with the set of shipped default names should equal 'names of defaults actually filtered out'.
- Currently: `disabled_defaults ⊋ (disabled_defaults ∩ default_names)` whenever the user has a typo.

Test 16 catches the warning (good) and that the real name is filtered (good), but does NOT assert what `configuration.disabled_defaults` contains in the output — so this asymmetry slips through.

Ratings: Encap 6/10 · Invar-Expr 4/10 · Invar-Useful 7/10 · Invar-Enforce 4/10

Suggested minimal improvement: split the field. Either (a) filter `disabled_defaults` to the intersection with `$default_agent_names` and add a sibling `unknown_disabled` field surfacing the typos for audit-tool consumption, or (b) keep `disabled_defaults` as raw input but add `disabled_defaults_applied` as the actually-filtered subset. (a) is cleaner — it makes `disabled_defaults` mean what its name suggests. Add an assertion to Test 16: `.configuration.disabled_defaults == ["pr-test-analyzer"]` (after filtering) and `.configuration.unknown_disabled == ["nonexistent-reviewer"]`. This closes the warning-only-on-stderr asymmetry and makes the invariant machine-checkable.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -153,47 +191,158 @@ parse_agent_file() {
...
-# Collect context sections (no dedup needed — all are passed through).
-# Build a JSON array of changed files for scoping
+# Build JSON array of changed files for scoping
 CHANGED_FILES_JSON=$(jq -n --arg files "$CHANGED_FILES" '$files | split("\n") | map(select(. != ""))')
 
-printf '%s\n' "$ALL_SECTIONS" | jq -s --argjson files "$CHANGED_FILES_JSON" '
+# Load defaults from plugins/pr-review-loop/agents/*.md.
+# Warn (don't fail) when zero defaults load — a broken plugin install would
+# leave the user with no specialist reviewers, and silent zero-defaults is
+# exactly the user-visible failure mode this check exists to surface.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


clarity: this 3-line WHY comment is overlong for a single-line warning whose own message ("plugin install may be incomplete") already conveys the diagnosis. The mid-comment phrase "silent zero-defaults is exactly the user-visible failure mode this check exists to surface" is meta-justification (explaining that the check needs explaining).

Verbose:
`# Load defaults from plugins/pr-review-loop/agents/*.md.
# Warn (don't fail) when zero defaults load — a broken plugin install would
# leave the user with no specialist reviewers, and silent zero-defaults is
# exactly the user-visible failure mode this check exists to surface.`

Terse:
`# Load defaults; warn (non-fatal) on zero — likely a broken plugin install.`

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -153,47 +191,158 @@ parse_agent_file() {
...
+    if [[ "$f_canon" != "$ROOT_AGENT_REVIEWERS_CANON" ]] && grep -q '^# Configuration[[:space:]]*$' "$f"; then
+        echo "Warning: # Configuration section found in $f — ignored (only the root AGENT-REVIEWERS.md's configuration is honored)" >&2
+    fi
+done
+
+# Warn on disabled names that don't match any default — likely typos.
+# This mirrors the strict validation we apply to overlap_acknowledged.reason:
+# the configuration schema asks the user to name defaults precisely, and a
+# typo like `disabled: ["pr-test-analyser"]` (British spelling) would
+# silently no-op without this check.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


clarity: 5-line preamble for a typo-detection warn. The "This mirrors the strict validation we apply to overlap_acknowledged.reason: the configuration schema asks the user to name defaults precisely" sentence is meta ("we are validating because we like validating"). The British-spelling example IS useful — keep it, drop the rest.

Verbose:
`# Warn on disabled names that don't match any default — likely typos.
# This mirrors the strict validation we apply to overlap_acknowledged.reason:
# the configuration schema asks the user to name defaults precisely, and a
# typo like \`disabled: ["pr-test-analyser"]\` (British spelling) would
# silently no-op without this check.`

Terse:
`# Warn on disabled names that don't match any default — typos like
# \`disabled: ["pr-test-analyser"]\` would otherwise silently no-op.`

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -153,47 +191,158 @@ parse_agent_file() {
...
+
+    # Disabled defaults (from configuration)
+    ($config.disabled // []) as $disabled_defaults |
+
+    # ---- 4. Effective defaults: not overridden, not disabled ----
+    # Precedence note: if a default appears in BOTH the disabled list AND is
+    # also overridden by a same-name user agent, the user agent still runs
+    # (as user-override). The default itself is filtered out by EITHER rule.
+    # In effect: the user has expressed both opt-out and replacement — both
+    # desires are satisfied.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


clarity: lines 303-304 ("In effect: the user has expressed both opt-out and replacement — both desires are satisfied") restate the preceding two lines ("the user agent still runs (as user-override). The default itself is filtered out by EITHER rule"). The "In effect" sentence is the same fact phrased from the user's POV. Drop it; the preceding two lines already convey the semantics.

Verbose:
`# Precedence note: if a default appears in BOTH the disabled list AND is
# also overridden by a same-name user agent, the user agent still runs
# (as user-override). The default itself is filtered out by EITHER rule.
# In effect: the user has expressed both opt-out and replacement — both
# desires are satisfied.`

Terse:
`# Precedence note: if a default is BOTH disabled and overridden by a same-name
# user agent, the user agent still runs (as user-override); the default is
# filtered out by EITHER rule.`

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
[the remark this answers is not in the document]
devonjones: Fixed - already addressed in round 1 (commit b2ba3da). Rewrote the frontmatter parser using portable substr-based parsing — no more gawk match() capture-group syntax. BSD awk on macOS now produces identical output.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -153,47 +184,121 @@ parse_agent_file() {
...
+    f_canon="$(realpath "$f" 2>/dev/null || echo "$f")"
+    if [[ "$f_canon" != "$ROOT_AGENT_REVIEWERS_CANON" ]] && grep -q '^# Configuration[[:space:]]*$' "$f"; then
+        echo "Warning: # Configuration section found in $f — ignored (only the root AGENT-REVIEWERS.md's configuration is honored)" >&2
+    fi
+done
+
+# Prepare input for jq:
+#   {sections: [...], defaults: [...], configuration: {...}, current_version: "..."}
+JQ_INPUT=$(jq -n \
+    --argjson sections "$([[ -z "$ALL_SECTIONS" ]] && echo "[]" || (printf '%s\n' "$ALL_SECTIONS" | jq -s '.'))" \
```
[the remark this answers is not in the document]
devonjones: Out of scope - tracked in devon-claude-skills-dxi. Current JSON sizes are ~45KB total (6 defaults + 4 user agents at typical prompt lengths), well under the 128KB Linux MAX_ARG_STRLEN. Defensive refactor to --slurpfile is filed as a focused follow-up so this PR can land.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
devonjones: Fixed - new tests/test-load-defaults.sh covers all three skip-with-warning paths (no frontmatter / unterminated frontmatter / empty or missing name) plus mixed-dir and empty-dir. 23 assertions, all passing. Total test surface across the two harnesses: 90 assertions.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - updated the documented output shape header to include 'description?' and clarified it's carried only on default entries (along with model? and color?, all from frontmatter).

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - collapsed to 'if ! JQ_PARSE_ERR=$(...)' form. Behavior-identical (if/! already suppresses set-e abort), no temporary set-state mutation, no intermediate variable.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_load_defaults.sh
devonjones: Fixed - inverted to 'if ! DEFAULT_AGENTS_DIR=$(cd ...)' form, removing the no-op ':' arm.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - added sibling field 'disabled_defaults_unknown' that captures only the typo'd names. The existing 'disabled_defaults' continues to echo raw user input for backwards compat. Test 16 now asserts both fields. JSON consumers can detect typos without parsing stderr.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - dropped the 3-line meta-justification; the warning message itself ('plugin install may be incomplete') conveys the diagnosis.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - trimmed to one line; kept the British-spelling example as the load-bearing illustration.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - dropped the 'In effect:' restatement. The two preceding lines convey the precedence.

PR #21, opened 2026-05-15
feat(pr-review-loop): ydy first cut — independent validator pipeline + model mirroring

devonjones:
## Summary

First cut of [devon-claude-skills-ydy](https://github.com/devonjones/devon-claude-skills/issues/ydy): per-finding validation step that runs between agent return and posting. Validators **mirror the flagger's model** — bug-class agents (silent-failure-hunter, pr-test-analyzer, code-simplifier) flag in opus and get opus validators; compliance agents (code-reviewer, comment-analyzer, type-design-analyzer) flag in sonnet and get sonnet validators.

Direct address of the explicit user-stated goal "not always using opus for reviewers" — half the validators end up sonnet (compliance) instead of blanket opus.

## Why

Catches the asymmetric-cost failure mode: a false-positive that gets *applied as a fix* introduces a real regression in once-correct code. Even when reviewers are 90%+ accurate, the dangerous 10% justifies a verification step.

## What ships

- **`# Configuration .independent_validator`** with `enabled` / `skip_for` / `uncertain_action` fields
- **Strict parser validation**: enum check on `uncertain_action`, type checks on `enabled` and `skip_for`, warning on unknown nested keys
- **discover-agents.sh extension** surfaces resolved validator config (defaults filled in) so the orchestrator reads it once at startup
- **SKILL.md additions**: full Independent Validator Pipeline section with the validator Task prompt template, model selection rule, configuration spec, telemetry shape; updated Spawning Summary, loop step 5, merge-readiness summary, and # Configuration field doc
- **Test surface**: 6 new scenarios in tests/test-discover-agents.sh covering all the validation paths

## Implementation notes

- Validator model selection: just **mirror the flagger's model**. No per-category map needed; the model itself encodes the tier (opus for bug-class agents, sonnet for compliance agents). Cleaner than the per-category mapping originally specced in 0np.
- Used explicit `has` checks in jq (not `//`) for `enabled` because `//` treats `false` as falsy and would mis-default an explicit `enabled: false` to true.
- Apostrophes in jq comments still break bash single-quoting (hit it again — fixed by rewording).

## Test results

- `tests/test-discover-agents.sh`: 84/84 (was 67; +17 new assertions for validator config)
- `tests/test-load-defaults.sh`: 23/23 (unchanged)
- Total: **107 assertions, all passing**

## Out of scope (separate tickets)

- Cross-family validator (Gemini validates Anthropic, etc.) — [devon-claude-skills-3m6](https://github.com/devonjones/devon-claude-skills/issues/3m6) Phase C
- Aggregate evaluation of validator effectiveness over a corpus — [devon-claude-skills-bl7](https://github.com/devonjones/devon-claude-skills/issues/bl7) (replay harness) + [devon-claude-skills-gx4](https://github.com/devonjones/devon-claude-skills/issues/gx4) (telemetry framework)

## Test plan

- [x] Local: both test harnesses 107/107 passing
- [x] Local: parser correctly rejects bad enums, types, and shapes
- [x] Local: explicit `enabled: false` honored (jq `has` check, not `//`)
- [ ] Dogfood the loop on this PR (round count expected to be small — bash + markdown changes only)


plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -97,4 +102,61 @@ if [[ -n "$INVALID_ENTRIES" ]]; then
...
+    exit 1
+fi
+
+# Validate skip_for is array of strings when present.
+IV_BAD_SKIP="$(echo "$RAW_JSON" | jq -r '
+    .independent_validator.skip_for // empty
+    | if type != "array" then "<not an array>"
+      elif any(.[]; type != "string") then "<contains non-string entries>"
+      else empty end
+')"
```
devonjones, line 167: 🤖 **Claude Code** (shell-script-reviewer):


[low] The skip_for diagnostic embeds the literal strings `<not an array>` / `<contains non-string entries>` from jq into the user-facing error. Functional, but reads slightly oddly ("Error: ... <not an array> — must be an array of strings"). Minor — consider branching in shell on a sentinel like `array` / `mixed` and producing two distinct messages. Not a blocker.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -675,10 +683,115 @@ Once the stale check passes (or `--skip-stale-check` is set), the loop emits a o
...
+When validation is disabled or partially skipped, the third line reflects that:
+
+```
+Validation: enabled (skip_for: code-simplifier).
+Validation: disabled.
+```
+
+### Independent Validator Pipeline
+
+After every agent reviewer returns its findings — and BEFORE those findings are posted as line comments — the loop runs an independent validator subagent per finding. The validator's job is to verify that the flagged issue is real, working from the diff context alone without seeing who flagged it or how confidently. Findings the validator rejects are dropped before posting.
```


Restatement: "working from the diff context alone without seeing who flagged it or how confidently" pre-states what's enumerated 6 lines below under "What the validator receives / does NOT receive". The trailing clause also reads awkwardly ("how confidently").

Verbose (current):
> The validator's job is to verify that the flagged issue is real, working from the diff context alone without seeing who flagged it or how confidently.

Terse:
> The validator verifies that the flagged issue is real. (Inputs/exclusions are spelled out below.)

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -675,10 +683,115 @@ Once the stale check passes (or `--skip-stale-check` is set), the loop emits a o
...
+
+This catches the asymmetric-cost failure mode: a false-positive that gets *applied as a fix* introduces a real regression in once-correct code. Even when agents are 90%+ accurate, the dangerous 10% justifies a verification step.
```


Second sentence is justifying throat-clearing — the asymmetric-cost framing in the first sentence already makes the case. "The dangerous 10%" is editorial.

Verbose (current):
> This catches the asymmetric-cost failure mode: a false-positive that gets *applied as a fix* introduces a real regression in once-correct code. Even when agents are 90%+ accurate, the dangerous 10% justifies a verification step.

Terse:
> This catches the asymmetric-cost failure mode: a false-positive applied as a fix introduces a regression in once-correct code.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -97,4 +102,61 @@ if [[ -n "$INVALID_ENTRIES" ]]; then
...
+# Validate enabled is boolean when present.
+IV_BAD_ENABLED="$(echo "$RAW_JSON" | jq -r '
+    .independent_validator.enabled // empty
+    | select(type != "boolean")
+    | tojson
+')"
+if [[ -n "$IV_BAD_ENABLED" ]]; then
+    echo "Error: # Configuration .independent_validator.enabled in $FILE must be a boolean (got $IV_BAD_ENABLED)" >&2
+    exit 1
+fi
```
devonjones, line 158: 🤖 **Claude Code** (shell-script-reviewer):


[low] The four IV_* validation steps each spawn a separate jq process on the same RAW_JSON. Functionally fine (RAW_JSON is small, runs once per discovery), but they could be consolidated into a single jq pipeline that emits a structured pass/fail + reason, then the shell branches on that. Pure perf/style nit — not a blocker, and the current form is arguably more readable.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -675,10 +683,115 @@ Once the stale check passes (or `--skip-stale-check` is set), the loop emits a o
...
+- **VALID** → post the finding through the normal flow (`post-line-comment.sh` / pending-review batching)
+- **INVALID** → drop silently from this round's posted findings; record in telemetry
+- **UNCERTAIN** → behavior depends on `# Configuration .independent_validator.uncertain_action`:
+  - `post_with_annotation` (default) — post with a `[validator: uncertain]` annotation
+  - `post_silently` — post normally
+  - `drop` — drop like INVALID
+
+#### Validator model selection
+
+The validator's model **mirrors the flagger's model.** Bug-class agents (silent-failure-hunter, pr-test-analyzer, code-simplifier; future security-reviewer / performance-reviewer via 9ao) flag in opus and get opus validators. Compliance agents (code-reviewer, comment-analyzer, type-design-analyzer) flag in sonnet and get sonnet validators. This avoids the failure mode of blanket-opus validation while keeping the verification step at appropriate strength.
```


Trailing clause restates what the preceding sentences just established. The opus-mirrors-opus / sonnet-mirrors-sonnet mapping IS the avoidance of blanket-opus and the appropriate-strength choice — saying so explicitly is throat-clearing.

Verbose (current):
> ... Compliance agents (...) flag in sonnet and get sonnet validators. This avoids the failure mode of blanket-opus validation while keeping the verification step at appropriate strength.

Terse: drop the second sentence entirely. The mirroring rule speaks for itself.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -675,10 +683,115 @@ Once the stale check passes (or `--skip-stale-check` is set), the loop emits a o
...
+    "uncertain_action": "post_with_annotation"
+  }
+}
+```
+
+- `enabled` (default `true`) — toggle validation on/off
+- `skip_for` (default `[]`) — list of flagger agent names whose findings bypass validation
+- `uncertain_action` (default `"post_with_annotation"`) — `post_with_annotation` | `post_silently` | `drop`
+
+The parser rejects unknown nested keys, non-boolean `enabled`, non-array `skip_for`, and `uncertain_action` outside the enum.
```


Inaccurate: this says the parser **rejects** unknown nested keys, but `_parse_configuration.sh` lines 118-126 only emit a **warning** for unknown nested keys (e.g. `enabld`) and continue parsing. Test 22 explicitly confirms this with `assert_exit "exit 0 (warning, not error)"`. The other three failure modes listed (non-boolean `enabled`, non-array `skip_for`, bad `uncertain_action` enum) are correctly rejected with exit 1, but unknown nested keys are not. Suggest: "The parser warns on unknown nested keys and rejects non-boolean `enabled`, non-array `skip_for`, and `uncertain_action` outside the enum."

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -675,10 +683,115 @@ Once the stale check passes (or `--skip-stale-check` is set), the loop emits a o
 ```
 Spawning 7 reviewers: 5 defaults + 1 user override (code-reviewer) + 1 user agent (pci-auditor).
```


Arithmetic mismatch in the example. Setup: 6 defaults total, pr-test-analyzer disabled, code-reviewer overridden by user, plus a user pci-auditor. That gives 4 unmodified defaults + 1 user-override (code-reviewer, fills the override slot) + 1 user agent (pci-auditor) = **6 reviewers**, not 7. (Compare with test 12 in tests/test-discover-agents.sh which uses an analogous setup and asserts `spawned_count == 6` with `default_count == 4, override_count == 1, user_count == 1`.) The line 686 follow-up ("3 bug-class agents, sonnet validators for 4 others") then also doesn't match: with code-reviewer overridden (user override defaults to sonnet), the actual split is 3 opus (silent-failure-hunter, pr-test-analyzer disabled... wait pr-test-analyzer was disabled, so 2 opus: silent-failure-hunter + code-simplifier) + 4 sonnet (comment-analyzer, type-design-analyzer, code-reviewer-override, pci-auditor) = 6. The example numbers are internally inconsistent. Suggest reworking with the actual derivation, e.g.: "6 reviewers: 4 defaults + 1 user override (code-reviewer) + 1 user agent (pci-auditor). Disabled defaults: pr-test-analyzer. Validation: enabled (opus validators for 2 bug-class agents, sonnet validators for 4 others)."

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -97,4 +102,61 @@ if [[ -n "$INVALID_ENTRIES" ]]; then
...
+')"
+if [[ -n "$IV_BAD_ACTION" ]]; then
+    echo "Error: # Configuration .independent_validator.uncertain_action in $FILE has invalid value '$IV_BAD_ACTION'" >&2
+    echo "Error:   Allowed values: post_with_annotation, post_silently, drop" >&2
+    exit 1
+fi
+
+# Validate enabled is boolean when present.
+IV_BAD_ENABLED="$(echo "$RAW_JSON" | jq -r '
+    .independent_validator.enabled // empty
```
devonjones: 🤖 **Claude Code** (code-reviewer):


**Bug: explicit `null` slips past validation for `enabled`, `skip_for`, and `uncertain_action`.** [confidence: high]

The pattern `.independent_validator.<field> // empty` uses jq's `//` operator, which treats both `null` and `false` as falsy. So when a user writes `{ "enabled": null }` (or `null` for any of the three nested fields), `null // empty` yields `empty`, the `select(...)` chain produces nothing, and validation passes. Reproduced:

```json
{ "independent_validator": { "enabled": null } }
```

Parser: exit 0, no error. Then `discover-agents.sh` emits `enabled: null` (because `has("enabled")` is true for null values), and the orchestrator's downstream truthiness check on `.enabled` will treat null as falsy — silently disabling the validator the user thought they were configuring.

Same hole exists for `skip_for: null` (line 152) and `uncertain_action: null` (line 130).

Suggested fix: switch each `<field> // empty` to `(if has("<field>") then .<field> else empty end)`, mirroring the deliberate `has`-based pattern in `discover-agents.sh` (which calls out this exact concern in its inline comment). That makes the parser reject `null` consistently with how it rejects strings/numbers.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -675,10 +683,115 @@ Once the stale check passes (or `--skip-stale-check` is set), the loop emits a o
...
 Disabled defaults: pr-test-analyzer.
+Validation: enabled (opus validators for 3 bug-class agents, sonnet validators for 4 others).
```
devonjones: 🤖 **Claude Code** (code-reviewer):


**Off-by-one in agent count.** [confidence: high]

`sonnet validators for 4 others` should be `3 others`. The default agent set is 6 total, split 3/3:
- opus (bug-class): silent-failure-hunter, pr-test-analyzer, code-simplifier
- sonnet (compliance): code-reviewer, comment-analyzer, type-design-analyzer

Verified by reading each agent's `model:` frontmatter under `plugins/pr-review-loop/agents/`. The text two paragraphs down (line 727) correctly lists three sonnet agents.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: 🤖 **Claude Code** (code-reviewer):


**Doc claim contradicts parser behavior.** [confidence: high]

This sentence says the parser `rejects unknown nested keys`, but the actual code (`_parse_configuration.sh` lines 118-126) only emits a `Warning:` and continues with exit 0 — and test 22 in `test-discover-agents.sh` explicitly asserts `exit 0 (warning, not error)` for the unknown-key case.

Suggest splitting the rejected vs warned cases, e.g.:

> The parser rejects non-boolean `enabled`, non-array `skip_for`, and `uncertain_action` outside the enum. Unknown nested keys produce a warning (so users can see typos like `enabld`) but do not fail the loop.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - switched to select(type == 'object' and has(...)) per Gemini's same finding (convergent independent confirmation - good signal). Test 22a (null enabled) and 22b (null skip_for) added as regression tests. The // empty bug now lives only in git history.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -97,4 +102,61 @@ if [[ -n "$INVALID_ENTRIES" ]]; then
...
+    | if type != "array" then "<not an array>"
+      elif any(.[]; type != "string") then "<contains non-string entries>"
+      else empty end
+')"
+if [[ -n "$IV_BAD_SKIP" ]]; then
+    echo "Error: # Configuration .independent_validator.skip_for in $FILE: $IV_BAD_SKIP — must be an array of strings" >&2
+    exit 1
+fi
+
 echo "$RAW_JSON" | jq -c .
```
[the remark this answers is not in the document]
devonjones: Fixed - both: switched to select(type == 'object' and has(...)) to avoid // empty swallowing null/false (code-reviewer found the same issue independently); applied printf '%s\n' "$RAW_JSON" everywhere we pipe to jq.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -73,12 +73,17 @@ fi
...
 UNKNOWN_KEYS="$(echo "$RAW_JSON" | jq -r '
-    [keys[] | select(. != "defaults_version_checked" and . != "disabled" and . != "overlap_acknowledged")]
+    [keys[] | select(
+        . != "defaults_version_checked"
+        and . != "disabled"
+        and . != "overlap_acknowledged"
+        and . != "independent_validator"
+    )]
     | join(", ")
 ')"
```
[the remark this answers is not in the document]
devonjones: Fixed - applied printf '%s\n' to all $RAW_JSON pipes in _parse_configuration.sh (10 sites). Error messages stay on echo per repo convention.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - corrected to '2 bug-class agents, sonnet validators for 4 others' to match the corrected 6-reviewer example. Added a parenthetical math walkthrough so the count is verifiable.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - corrected to '6 reviewers: 4 defaults + 1 user override + 1 user agent (pci-auditor)' with explicit math walkthrough below the example.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - rewrote to distinguish what the parser rejects (non-boolean enabled, non-array skip_for, bad uncertain_action enum, non-object independent_validator) from what it warns about without rejecting (unknown nested keys).

devonjones: Fixed - same change as PRRC_kwDOQsw1MM7By2ki / comment 3251333410. Convergent finding from comment-analyzer and code-reviewer.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - removed the angle-bracket sentinel strings from the user-facing error. Now reads 'must be an array (got string)' or 'must be an array of strings (got non-string entry)'. Test 21's assertion updated to match.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Won't fix - you noted current form is 'arguably more readable' yourself, and I agree. Four focused jq invocations are clearer than one combined invocation here because each validates a distinct field with distinct error semantics. Keeping the readability trade-off where it is.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - dropped the pre-statement that duplicated the explicit 'what the validator does/does not see' enumeration 6 lines below. 'How confidently' phrasing also gone.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - dropped the 'even when agents are 90%+ accurate, the dangerous 10% justifies' sentence. The asymmetric-cost framing in the preceding sentence already lands the point.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - dropped the 'avoids blanket-opus / appropriate strength' restatement. The opus-mirrors-opus / sonnet-mirrors-sonnet rule already implies that framing.

PR #24, opened 2026-05-20
feat(pr-review-loop): backport qrc + skj — priority mapping + quality-weighted exit + Hard Round Ceiling

devonjones:
## Summary

Backports two SKILL.md sections from `valid-eval-skills` (downstream) to finish the **jq0 backport epic** alongside its+2pp (PR #23, merged as `79d2f8f`):

- **qrc**: Priority to Exit-Condition Mapping table + Gemini `![medium]` classification heuristics
- **skj**: quality-weighted exit condition + Hard Round Ceiling (Circuit Breaker)

Dogfooded over 4 rounds — see commit history. Round 3 caught and fixed three real logic gaps; round 4 propagated terminology consistently; round 5 surfaced a Gemini self-contradiction (rejected with P2→P3 reclassification).

## Why

Two gaps in the existing Stopping Heuristics that surfaced during the PR #23 dogfood:

1. The Supported Review Bots table listed the priority **formats** but not what to do with them — so the loop treated "Won't fix on a real bug" as resolution when it shouldn't.
2. The ONE MORE LOOP Rule treated all feedback symmetrically — any fix resets the counter — so even one nitpick polish kept extending the loop. Quality-weighted exit distinguishes P1/P2 (resets) from P3 (counts toward exit condition b).

The Hard Round Ceiling is a circuit breaker for the case where neither stopping heuristic nor quality-weighted exit fires — stop and ask the user.

### qrc — Priority to Exit-Condition Mapping (new subsection)
Inserted after the Supported Review Bots table:
- Table mapping each source label (Gemini, Cursor, Claude, agents) to P1/P2/P3 with a "blocks quality-weighted exit?" column
- "Won't fix on a P1/P2 finding does NOT resolve it" rule
- `![medium]` classification heuristics — correctness/security/breaking → P2 blocking; style/prose → P3 non-blocking; ambiguous → default P2 first round, downgrade only if fix is a judgment call

### skj — Hard Round Ceiling + quality-weighted exit (rewritten)
- New `### Hard Round Ceiling (Circuit Breaker)` subsection: **pure circuit breaker on 7 total rounds**. Earlier nitpick-based termination is handled by the quality-weighted exit + ONE MORE LOOP combination, not by the ceiling.
[...]

PR #25, opened 2026-05-20
feat(pr-review-loop): language-aware default AGENT-REVIEWERS.md templates with copy-and-extend offer (98b)

devonjones:
## Summary

Closes the **98b epic**: language-aware default AGENT-REVIEWERS.md templates. When a user runs the pr-review-loop on a repo with no AGENT-REVIEWERS.md, the loop now offers to install an opinionated reviewer pack for whatever language(s) it detects in the repo (golang or python today; more to come). After install, it can also propose project-specific reviewers tailored to what it sees in the codebase.

Bumps pr-review-loop plugin **1.3.0 → 1.4.0** (feature add, SemVer minor).

Six commits across five tranches (see "Commits" below). All tracked via beads child tickets under 98b.

### Templates (`plugins/pr-review-loop/agent-reviewer-templates/`)

Two opinionated language packs authored by the user (reshape-only edits in this PR per Option A reconciliation with the existing AGENT-REVIEWERS.md spec):

| Template | Reviewers | Disables (subsumed by pack) | Overlap-acknowledges |
|----------|-----------|------------------------------|----------------------|
| `golang.md` (902 lines) | 8: test-coverage, error-handling, complexity, concurrency, resource-leak, external-process, dead-code, clarity | silent-failure-hunter, comment-analyzer, code-simplifier | pr-test-analyzer |
| `python.md` (1674 lines) | 13: test-coverage, error-handling, logging, complexity, concurrency, resource-leak, external-process, dead-code, import, importlib-resources, dataclass-decorator, typing-consistency, clarity | (same 3) | (same) |

Reshape applied per Option A:
- H1 rename: `# Code Review Agents — <Lang>` → `# Agents` (the H1 the parser recognises)
- Pack-level metadata sub-sections (How the pack runs / Tooling assumed in CI / Severity convention / Output format / Deferring with beads) moved under `# Guidelines`
- Priority shift: P0/P1/P2 → P1/P2/P3 (atomic, via placeholder substitution to avoid double-shift). Aligns with qrc Priority Mapping shipped in PR #24

### Scripts (`plugins/pr-review-loop/skills/pr-review-loop/scripts/`)

**`detect-language.sh`** — walks repo for known manifests (`go.mod`, `pyproject.toml`, `setup.py`, `requirements.txt`), emits a JSON array of `{language, subtree, manifest}` pairs. Deterministic, no LLM. Prunes common ignored dirs (`.git`, `node_modules`, `vendor`, etc.). Handles single-language, per-subtree polyglot, same-directory polyglot, Go workspaces.

**`install-template.sh`** — takes `<lang>:<subtree>` pairs, installs templates per-subtree, writes a merged `# Configuration` to the root AGENT-REVIEWERS.md. Transactional refusal (validates ALL targets before writing anything). Smart `overlap_acknowledged` merge: same agent + same `overlaps_with` → concatenate `reason` strings with ` | `; same agent + different `overlaps_with` → fatal conflict.

### discover-agents.sh integration

When NO AGENT-REVIEWERS.md exists anywhere (root or via walk-up from changed files), `discover-agents.sh` now calls `detect-language.sh` and surfaces a `language_detection` block in its JSON output. Otherwise `null` (suppressed).

```json
"language_detection": {
  "matched": [
    {"language": "golang", "subtree": "backend", "manifest": "go.mod", "template": "golang.md"},
    {"language": "python", "subtree": "frontend", "manifest": "pyproject.toml", "template": "python.md"}
  ],
  "same_directory_polyglot": false
}
```

### SKILL.md docs

Two new sections in the Agent Reviewers area:

- **`### Language Template Offer`** — pre-loop step 2; documents prompt shapes for single-match / per-subtree polyglot / same-directory polyglot; install invocation; decline behaviour
- **`### Project-Specific Reviewer Recommendations`** — post-install opt-in flow that proposes 3-7 repo-specific reviewers tailored to the codebase

### Design pivot: tranche d (recommendations)

Original sketch was "enter Claude Code's plan mode and let the user review the proposed plan via ExitPlanMode". Verified via the claude-code-guide agent that this **doesn't work**: `EnterPlanMode` exists but `ExitPlanMode` returns no structured data — there's no programmatic way for a skill to read back what the user approved. Pivoted to `AskUserQuestion` with `multiSelect: true` (capped at 4 per call; 5-7 proposals batched with "(N of M)" header).

Trade-off documented in SKILL.md: no "modify before accept" UX — user accepts or rejects each proposal as written; manual edits to AGENT-REVIEWERS.md happen after.

### Multi-language strategy

Chosen: **per-subtree install** using the existing hierarchical scoping infrastructure. The 8 same-named reviewers between golang.md and python.md don't collide because they live in different files (e.g., `/backend/AGENT-REVIEWERS.md` for Go, `/frontend/AGENT-REVIEWERS.md` for Python). Same-directory polyglot is the rare-edge fallback → user pick-one prompt.

## Commits

| Commit | Tranche | Beads | What |
|--------|---------|-------|------|
| `fa6d618` | (foundation) | 98b | Reshaped golang.md + python.md templates → `agent-reviewer-templates/` |
| `6b92fa6` | a | 6eo | `detect-language.sh` + `install-template.sh` + 42 test assertions |
| `c3d266c` | b | njz | `discover-agents.sh` integration + 14 new test assertions |
| `f1a1f8f` | c | kxi | SKILL.md Language Template Offer section |
| `c808a81` | d | dbs | SKILL.md Project-Specific Reviewer Recommendations section (AskUserQuestion pivot) |
| `0f784ac` | bump | — | plugin 1.3.0 → 1.4.0 + AGENT-REVIEWERS.md pin |

All beads child tickets closed: 6eo, njz, kxi, dbs. 98b epic closed.

## Test plan

- [x] `tests/test-discover-agents.sh` 102/102 (was 88; +14 for language_detection)
- [x] `tests/test-load-defaults.sh` 23/23
- [x] `tests/test-language-templates.sh` 42/42 (new — covers detect-language.sh + install-template.sh)
- [x] **Total: 167 assertions passing**
- [ ] Dogfood — run the loop on this PR (validates the new pre-loop step 2 doesn't fire spuriously on a repo that already has AGENT-REVIEWERS.md)

## Migration

Existing user repos pinned to 1.3.0 will see the stale-pin warning on next loop invocation. Migration paths:
1. Re-run with `--skip-stale-check` and bump the pin to 1.4.0 manually (today)
2. `/pr-review-loop:audit-agents` (q2h) — not yet built; tracked separately


plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -357,11 +357,14 @@ The review loop should **not** override branch protections or bypass repo-define
 1. **Discover and merge agents.** Run `scripts/discover-agents.sh <PR>` once. Its output includes:
    - The full merged agent list (defaults + user agents per C+E semantics)
    - The `configuration` block with `stale_pin`, `defaults_version_checked`, `current_plugin_version`, counts
-2. **Check the stale pin.** If `configuration.stale_pin` is `true` AND `--skip-stale-check` was not passed:
+   - The `language_detection` block (`null` when the repo already has any `AGENT-REVIEWERS.md`; otherwise `{matched: [...], same_directory_polyglot: bool}`)
+2. **Offer language templates** (only when `language_detection != null` AND `language_detection.matched` is non-empty). See "Language Template Offer" below for the prompt shape, the `install-template.sh` invocation, and the same-directory-polyglot fallback. If the user accepts, the install runs before round 1 and the loop continues with the newly-installed agents discoverable in subsequent steps. If the user declines, record the decline for the rest of this loop invocation so the offer doesn't re-fire — then continue to step 3 with only the 6 baked-in defaults.
```


**Verbose step 2 summary.** The last two sentences (`If the user accepts … discoverable in subsequent steps. If the user declines … 6 baked-in defaults.`) restate in detail what the "Language Template Offer" and "When the user declines" sub-sections already cover. The forward reference is enough.

Suggested trim:
```
2. **Offer language templates** (only when `language_detection != null` AND `language_detection.matched` is non-empty). See "Language Template Offer" below.
```

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
 - **`overlap_acknowledged`** — map from a user agent name to `{ overlaps_with, reason }`. Both agents continue to spawn; this entry documents intentional duplication so the audit tool doesn't recommend renaming. **`reason` is REQUIRED** — the parser rejects entries without it, so future readers see why both agents are intentionally running.
 - **`independent_validator`** — controls the per-finding validation step (see "Independent Validator Pipeline" below). All three nested fields are optional; defaults are `enabled: true`, `skip_for: []`, `uncertain_action: "post_with_annotation"`.
 
+### Language Template Offer
+
+When the repo has no `AGENT-REVIEWERS.md` anywhere (root or in any subdirectory walked from changed files), `discover-agents.sh` calls `detect-language.sh` and surfaces a `language_detection` block in its output. Pre-loop step 2 reads that block and, when non-empty, offers to install language-specific reviewer templates before the loop runs.
```


**Opening paragraph duplicates step 2 and "When the offer fires".** The paragraph explains when `discover-agents.sh` calls `detect-language.sh`, what the `language_detection` block contains, and when the offer fires — all of which are stated in pre-loop step 2 (line 360-361) and spelled out again in the sub-section directly below (lines 781-786).

The only new information here is the location of templates. Suggest replacing with a single orientation sentence:

```
When no `AGENT-REVIEWERS.md` exists anywhere in the repo, `discover-agents.sh` populates a `language_detection` block; if it has matches, the loop offers to install a language-specific reviewer template before round 1. Templates live under `plugins/pr-review-loop/agent-reviewer-templates/`.
```

Then merge the template-description sentence from line 777 here as well, dropping the duplicated firing conditions.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+The plugin ships templates under `plugins/pr-review-loop/agent-reviewer-templates/`. Initial set: `golang.md`, `python.md`. Templates are opinionated bundles — each disables the baked-in defaults it explicitly subsumes (e.g., `silent-failure-hunter` is disabled by both packs because they ship richer `error-handling-reviewer` agents) and acknowledges legitimate overlaps with `pr-test-analyzer`.
+
+#### When the offer fires
+
+Only when ALL of these hold:
+- `language_detection != null` (no AGENT-REVIEWERS.md exists at root OR via walk-up from changed files)
+- `language_detection.matched` is non-empty (at least one manifest detected)
+- The user has not declined the offer earlier in this loop invocation
+
+If any condition fails, skip the offer silently and continue to step 3 (stale-pin check).
```


**Restates the negative of the preceding list.** "If any condition fails, skip the offer silently and continue to step 3 (stale-pin check)." is the logical inverse of "Only when ALL of these hold" — readers don't need it spelled out. Drop the line.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+# Per-subtree polyglot — install all
+scripts/install-template.sh golang:backend python:frontend
+
+# Same-directory polyglot — user picked golang
+scripts/install-template.sh golang:.
+```
+
+The script is transactional — it validates ALL target paths before writing anything and refuses if any `AGENT-REVIEWERS.md` already exists. Successful install bumps `defaults_version_checked` to the current plugin version, so the subsequent stale-pin check (step 3) passes by construction.
+
+#### After install: offer plan mode
```


**Heading says "plan mode" but the section explicitly rejects plan mode.** The "Design rationale" sub-section (line 873) explains that plan mode doesn't work and the actual mechanism is `AskUserQuestion`. The heading should match what the section actually does.

Suggested rename: `#### After install: offer project-specific reviewer recommendations`

This also removes the need for the last sentence of line 861 ("On accept, enter Claude Code's plan mode and follow…") to mention plan mode — it can just say "follow the 'Project-Specific Reviewer Recommendations' section below."

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+Template installed. Want recommendations for repo-specific reviewers
+based on the codebase (e.g., migration-safety-reviewer for DB-heavy
+repos, prompt-injection-reviewer for LLM-heavy repos)? [y/N]
+```
+
+On accept, enter Claude Code's plan mode and follow the "Project-Specific Reviewer Recommendations" section below (tranche d). On decline, proceed to step 3 with just the template install in place.
+
+#### When the user declines the install offer
+
+Don't re-prompt within the same loop invocation. Track the decline in TodoWrite (or equivalent ephemeral state) so a fresh loop invocation re-fires the offer if conditions still hold (i.e., the user can change their mind on the next PR). Continue to step 3 with only the 6 baked-in defaults.
```


**Parenthetical filler.** "(i.e., the user can change their mind on the next PR)" explains what "fresh loop invocation" already implies. Drop it.

Before: `so a fresh loop invocation re-fires the offer if conditions still hold (i.e., the user can change their mind on the next PR).`
After: `so a fresh loop invocation re-fires the offer if conditions still hold.`

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+
+### Project-Specific Reviewer Recommendations
+
+After `install-template.sh` succeeds and the user accepts the post-install "want recommendations?" prompt, the loop spawns a focused scan of the repo and proposes 3-7 additional reviewers tailored to what it finds (e.g., `migration-safety-reviewer` for a repo with a `/db/migrations/` tree, `prompt-injection-reviewer` for a repo with LLM API calls in the diff).
+
+This step is OPT-IN — skip it entirely if the user declined.
```


**States the obvious.** "This step is OPT-IN — skip it entirely if the user declined." The preceding sentence already conditions entry on the user accepting the prompt. An agent won't reach this section if the user declined. Drop the line.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+
+The scan that informs proposals is **the orchestrator's call** — there's no fixed list of files to read. Use these guidelines to set scan depth:
+
+| Always read (cheap, high-signal) | Optional (depth-dependent) |
+|----------------------------------|----------------------------|
+| `README.md` at root              | `.beads/issues.jsonl` recent open issues |
+| `CLAUDE.md` at root              | Top 20 commit messages (`git log --oneline -20`) |
+| Top-level dependency manifests (names, not lockfiles) | Sampled source files |
+
+**Scan budget**: keep the total read at ≤10k tokens of repo content. Stop scanning when you have enough signal to propose 3-7 reviewers — more scanning past that point produces overlapping proposals without adding signal.
```


**Rationale clause doubles sentence length without adding instruction.** "Stop scanning when you have enough signal to propose 3-7 reviewers" is the actionable rule. The trailing clause "more scanning past that point produces overlapping proposals without adding signal" explains why but doesn't change what the agent does.

Before: `Stop scanning when you have enough signal to propose 3-7 reviewers — more scanning past that point produces overlapping proposals without adding signal.`
After: `Stop scanning when you have enough signal to propose 3-7 reviewers.`

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md


**Stale cross-reference: heading says "offer plan mode" but the feature uses AskUserQuestion**

The heading `#### After install: offer plan mode` and the instruction on line 861 (`On accept, enter Claude Code's plan mode`) directly contradict the `#### Design rationale: AskUserQuestion, not plan mode` section a few lines below, which explains why plan mode was rejected.

The design pivot (recorded in beads issue `dbs`) explicitly replaced plan mode with `AskUserQuestion` with `multiSelect: true`. The heading and the sentence on line 861 were not updated to reflect this.

**Fix:** rename the heading and rewrite line 861. For example:

- Heading: `#### After install: offer repo-specific reviewer recommendations`
- Line 861: `On accept, follow the "Project-Specific Reviewer Recommendations" section below to propose and install repo-tailored reviewers via AskUserQuestion. On decline, proceed to step 3 with just the template install in place.`

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+
+**Source-file sampling**: for repos with <100 source files total, sample 5-10. For larger repos, cap at ~20. Bias toward larger / older files (more likely to encode patterns the team cares about). Skip generated code, tests, and vendored dependencies.
+
+**Don't read the full PR diff** — the PR is just one slice of the repo; proposals should serve the repo broadly, not be PR-specific.
+
+#### Proposal generation: spawn as a Task
+
+Use the Task tool with a focused prompt rather than inlining the scan in the main loop conversation. Keeps main-context tokens low and lets the scan be the only thing the agent thinks about.
```


**Second clause restates the first.** "Keeps main-context tokens low and lets the scan be the only thing the agent thinks about." The second clause ("lets the scan be the only thing") is another way of saying low context usage. One is enough.

Before: `Keeps main-context tokens low and lets the scan be the only thing the agent thinks about.`
After: `Keeps main-context tokens low.`

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+
+Once `install-template.sh` succeeds, offer a second prompt for repo-tailored reviewer recommendations:
+
+```
+Template installed. Want recommendations for repo-specific reviewers
+based on the codebase (e.g., migration-safety-reviewer for DB-heavy
+repos, prompt-injection-reviewer for LLM-heavy repos)? [y/N]
+```
+
+On accept, enter Claude Code's plan mode and follow the "Project-Specific Reviewer Recommendations" section below (tranche d). On decline, proceed to step 3 with just the template install in place.
```


**Same stale reference: instruction says "enter Claude Code's plan mode" after the design pivot to AskUserQuestion**

This sentence tells the executing agent to enter plan mode. The `#### Design rationale: AskUserQuestion, not plan mode` section immediately below this paragraph explains that plan mode was evaluated and found to not work (`ExitPlanMode returns no structured data`). The instruction here was never updated.

An agent following this SKILL.md literally would try to enter plan mode, discover it doesn't work (as documented 12 lines later), and have no fallback instruction.

This is the same bug as the line-851 heading comment; both need the same fix.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+
+```markdown
+## <proposal.name>
+
+<proposal.focus>
+
+<!-- Proposed during install-template setup; rationale: <proposal.rationale> -->
+```
+
+Append as a new `## ` block under `# Agents` in the target file. Preserve any existing agents. The rationale comment is informational — future audit-agents runs (q2h) can use it to explain why a reviewer is present.
```


**Non-actionable commentary.** "The rationale comment is informational — future audit-agents runs (q2h) can use it to explain why a reviewer is present." The comment template above already shows what to write; this sentence doesn't tell the agent to do anything different. Drop it.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,238 @@ Fields:
...
+
+#### After writing
+
+Emit a one-line summary to the spawning summary surface:
+
+```
+Installed 3 project-specific reviewer(s): migration-safety-reviewer (/db/), prompt-injection-reviewer (/), auth-bypass-reviewer (/).
+```
+
+No re-run of `discover-agents.sh` needed — the loop is about to proceed to round 1, and it picks up the new files on its next walk.
```


**Second clause restates the obvious.** "the loop is about to proceed to round 1, and it picks up the new files on its next walk" — the second clause is an implicit consequence of proceeding to round 1. Drop it.

Before: `No re-run of `discover-agents.sh` needed — the loop is about to proceed to round 1, and it picks up the new files on its next walk.`
After: `No re-run of `discover-agents.sh` needed — the loop proceeds to round 1 next.`

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -0,0 +1,127 @@
...
+#
+# Deduplication: when multiple manifests for the same language live in
+# the same subtree (e.g., pyproject.toml AND setup.py), the canonical
+# one wins. Priority within Python is pyproject.toml > setup.py >
+# requirements.txt.
+#
+# Usage:
+#   detect-language.sh [--repo-root <path>]
+
+set -euo pipefail
```
devonjones, line 28: 🤖 **Claude Code** (shell-script-reviewer):


All three new scripts (`detect-language.sh`, `install-template.sh`, `test-language-templates.sh`) and the modifications to `discover-agents.sh` pass all shell quality checks:

- `set -euo pipefail` present in all new scripts
- All error messages route to stderr (`>&2`)
- Exit codes are meaningful: non-zero for every failure path (exit 1 for refusals, exit 2 for bad input), zero on success
- Variables that could contain spaces or special characters are quoted throughout (e.g., `"$relpath"`, `"$subtree"`, `"$template_path"`, `"$target"`)

No defects found.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md


**Internal implementation term "tranche d" leaked into agent-facing instruction**

The phrase `(tranche d)` is an internal planning label from the beads issue breakdown — it has no meaning to an agent executing this SKILL.md. An agent reading this instruction will not know what "tranche d" means and cannot use the label to do anything. It should be removed.

devonjones: 🤖 **Claude Code** (code-reviewer):


**Agent-clarity bug: contradictory "plan mode" instruction**

Line 861 tells the agent: *"On accept, enter Claude Code's plan mode and follow the 'Project-Specific Reviewer Recommendations' section."* The section heading above it (line 851) is titled **"After install: offer plan mode"**.

Twelve lines later, line 875 explicitly says: **"The original sketch for this flow used Claude Code's plan mode. It doesn't work."** and prescribes `AskUserQuestion` instead.

Guideline: *"SKILL.md and `references/*.md` files are agent instruction documents — evaluate for agent clarity and accuracy."*

An agent reading top-to-bottom will reach line 861 first, attempt to enter plan mode, and only encounter the correction after the fact. This is a live contradiction in the instruction path.

**Fix**: update line 861 to say "follow the 'Project-Specific Reviewer Recommendations' section below (tranche d)" (drop the "enter Claude Code's plan mode" clause), and retitle the section heading at line 851 from "After install: offer plan mode" to "After install: offer repo-specific reviewer recommendations" (or similar).

Confidence: 95% — the contradiction is textual and unambiguous.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - collapsed step 2 to a single sentence + forward reference; accept/decline detail now lives only in the dedicated 'Language Template Offer' sub-section.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - merged the two opening paragraphs into a single orientation paragraph; dropped the duplicate firing-condition description.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - dropped the trailing 'If any condition fails, skip silently' line. The 'Only when ALL of these hold' framing above carries the meaning.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - renamed heading from 'After install: offer plan mode' to 'After install: offer repo-tailored recommendations' AND rewrote the body to call out AskUserQuestion. Convergent finding with code-reviewer and comment-analyzer on this same line - tranche d pivoted away from plan mode but I forgot to scrub the tranche c text. Caught by 4 reviewers.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - dropped the '(i.e., the user can change their mind on the next PR)' parenthetical. 'Fresh loop invocation' already implies it.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - dropped the 'This step is OPT-IN' line. The preceding 'the user accepts the post-install want-recommendations prompt' carries the opt-in semantic.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - tightened to 'Stop scanning once you have enough signal to propose 3-7 reviewers.'

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - rewrote the section heading and body to call AskUserQuestion explicitly, scrubbing all plan-mode references. Also dropped the '(tranche d)' planning jargon. Convergent finding with clarity-reviewer and code-reviewer.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -230,6 +239,38 @@ else
...
+                    matched: ($matches | map(. + {template: (.language + ".md")})),
+                    same_directory_polyglot: (
+                        # True iff two matches share a subtree with different languages
+                        ($matches
+                            | group_by(.subtree)
+                            | map(select(length > 1 and ([.[].language] | unique | length > 1)))
+                            | length) > 0
+                    )
+                }
+            ')
```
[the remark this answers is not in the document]
devonjones: Fixed - rewrote the language_detection jq block to pass RAW_DETECT via env var + 'env.RAW_DETECT | fromjson' instead of --argjson. Same pattern applied to the JQ_INPUT block (which carries SECTIONS_JSON, DEFAULTS_JSON, etc. — those are more likely to grow large) and the UNKNOWN_DISABLED block. Three sites total, all use env-var pattern now.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -258,14 +299,16 @@ if [[ -n "$UNKNOWN_DISABLED" ]]; then
...
 JQ_INPUT=$(jq -n \
     --argjson sections "$([[ -z "$ALL_SECTIONS" ]] && echo "[]" || (printf '%s\n' "$ALL_SECTIONS" | jq -s '.'))" \
     --argjson defaults "$DEFAULTS_JSON" \
     --argjson configuration "$CONFIGURATION_JSON" \
     --argjson files "$CHANGED_FILES_JSON" \
     --arg current_version "$CURRENT_PLUGIN_VERSION" \
-    '{sections: $sections, defaults: $defaults, configuration: $configuration, files: $files, current_version: $current_version}'
+    --argjson language_detection "$LANGUAGE_DETECTION_JSON" \
+    '{sections: $sections, defaults: $defaults, configuration: $configuration, files: $files, current_version: $current_version, language_detection: $language_detection}'
 )
```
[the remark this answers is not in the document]
devonjones: Fixed - this is the JQ_INPUT block, the most important of the three sites you flagged. SECTIONS_JSON in particular grows with the number of user agents in AGENT-REVIEWERS.md. Now passed via env + 'env.X | fromjson'. The current_version is small and stays as --arg.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - tightened to 'Keeps main-context tokens low.' The 'lets the scan be the only thing' clause was just restating it.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed in the same edit as the 851 fixes - dropped the '(tranche d)' jargon when rewriting the plan-mode reference. The flow now points to AskUserQuestion directly.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - tightened the rationale-comment line to 'supports future audit-agents (q2h) recommendations'. Drops the redundant 'informational' framing.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - tightened to 'No re-run of discover-agents.sh needed — round 1 picks up the new files naturally.' Dropped the 'about to proceed to round 1' redundancy with 'picks up on its next walk'.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
devonjones: Acknowledged - clean PASS on all five files in scope. Thanks for the line-by-line check; the explicit confirmation that exit codes / stderr routing / quoting are all conformant is useful as a per-PR audit trail.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed - the heading + body at line 851/861 both rewritten to remove plan-mode references. The post-install offer now reads 'After install: offer repo-tailored recommendations' and the body explicitly cites AskUserQuestion (multi-select). The '(tranche d)' jargon is also gone. 4 reviewers converged on this finding — strong signal that the touch-it-you-own-it discipline caught a real bug.

devonjones: Fixed in commit d4e8761 - this is a convergent finding (clarity-reviewer L851 + comment-analyzer L851/L861 + this code-reviewer L861 all flagged the same plan-mode contradiction). Missed your specific comment ID in my round-1 BATCH POINT because it posted after my C1 background poll returned. Fix applied: heading is now 'After install: offer repo-tailored recommendations' and the body calls AskUserQuestion explicitly. 4-way convergence — strong validation that touch-it-you-own-it works.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -767,6 +770,232 @@ Fields:
...
+# Per-subtree polyglot — install all
+scripts/install-template.sh golang:backend python:frontend
+
+# Same-directory polyglot — user picked golang
+scripts/install-template.sh golang:.
+```
+
+The script is transactional — it validates ALL target paths before writing anything and refuses if any `AGENT-REVIEWERS.md` already exists. Successful install bumps `defaults_version_checked` to the current plugin version, so the subsequent stale-pin check (step 3) passes by construction.
+
+#### After install: offer repo-tailored recommendations
```
devonjones, line 847: 🤖 **Claude Code** (comment-analyzer):


**Round 2 final verification: PASS — no new issues introduced**

All three round-1 fix commits verified clean:

**1. Heading rename: no stale orphans**
`grep -rn 'After install: offer plan mode'` → 0 hits anywhere in the repo. The old heading is gone. The new heading (`After install: offer repo-tailored recommendations`) appears only at SKILL.md:847, which is correct.

The two surviving `plan mode` occurrences in SKILL.md (lines 857 and 867) are both in the intentional design-rationale section — they exist to explain *why* plan mode was rejected, not to instruct an agent to use it. No contradiction with the new heading.

**2. Tranche-jargon: no orphan references**
`grep -rn 'tranche [a-z]'` → one hit: `tests/test-discover-agents.sh:636` — a test comment labeling (`# Test 23–26: language_detection block (98b tranche b)`). This is internal test scaffolding, not agent-facing instruction, and predates the fix commits. Not a problem.

Zero hits in SKILL.md, discover-agents.sh, or any other non-test file.

**3. env-var refactor: semantically equivalent**
All three jq invocations were refactored from `--argjson $VAR` to `VAR=$VAR jq ... env.VAR | fromjson`. The output objects are identical field-for-field:
- `LANGUAGE_DETECTION_JSON`: same `{matched, same_directory_polyglot}` shape
- `UNKNOWN_DISABLED`: same `map(.name)` + disabled subtraction + `join(", ")` pipeline
- `JQ_INPUT`: same 6-field object; `current_version` correctly stays as `--arg` (plain string, not JSON — no `fromjson` needed and none added)

The 102/102 test pass in the commit message independently confirms behavioral equivalence. Logic review agrees.

**Verdict**: fixes are complete, correct, and introduce no new cross-references or semantic drift. All 3 prior threads appropriately resolved.

devonjones: Acknowledged - thorough verification across all three round-1 fix dimensions (heading rename, tranche-jargon scrub, env-var refactor semantics). The PASS comment is exactly the audit-trail signal we want at this stage of the loop.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -0,0 +1,127 @@
...
+for d in "${skip_dirs[@]}"; do
+    prune_args+=(-name "$d" -prune -o)
+done
+
+# Build the -name expression for the find search.
+name_args=()
+for m in "${all_manifests[@]}"; do
+    name_args+=(-name "$m" -o)
+done
+unset 'name_args[${#name_args[@]}-1]'
```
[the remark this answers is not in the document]
devonjones: Fixed - replaced the unset trick with conditional '-o' insertion during the loop. Semantically equivalent for our case (always-non-empty), graceful for the empty edge. Also avoids the set -u risk with arithmetic indexing.

plugins/pr-review-loop/skills/pr-review-loop/scripts/install-template.sh
```diff
@@ -0,0 +1,285 @@
...
+        echo "Error: bad pair '$pair' — expected <lang>:<subtree>" >&2
+        exit 2
+    fi
+    lang="${pair%%:*}"
+    subtree="${pair#*:}"
+
+    if [[ -z "$lang" || -z "$subtree" ]]; then
+        echo "Error: bad pair '$pair' — language and subtree both required" >&2
+        exit 2
+    fi
```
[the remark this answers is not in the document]
devonjones: Fixed - added explicit path-traversal guard before any path is composed. Classified P2 (security) per the qrc Priority Mapping; counter resets per the exit logic. Guards: lang rejects '/' or leading '.', subtree rejects '..' segments (leading, embedded, trailing) and absolute paths. Added 4 test cases (I_T1-I_T4) covering each variant. The previous reliance on 'subtree must be an existing directory' wasn't enough — 'sub/../../etc' has every segment existing.

plugins/pr-review-loop/skills/pr-review-loop/scripts/install-template.sh
```diff
@@ -0,0 +1,285 @@
...
+        /^# / {
+            if (in_section) { in_section = 0 }
+            heading = substr($0, 3)
+            # Trim trailing whitespace
+            sub(/[[:space:]]+$/, "", heading)
+            if (heading == section) { in_section = 1; next }
+        }
+        in_section { print }
+    ' "$file"
+}
```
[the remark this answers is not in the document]
devonjones: Fixed - extract_section() now matches '#[[:space:]]+' and uses sub() to strip the prefix instead of substr($0, 3). A template with '#  Agents' (two spaces) or '#\tAgents' would have silently parsed as an unknown section under the old regex; now handled.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -0,0 +1,129 @@
...
+for d in "${skip_dirs[@]}"; do
+    prune_args+=(-name "$d" -prune -o)
+done
+
+# Build the -name expression for the find search. Conditionally insert
+# `-o` between terms rather than appending then unsetting the trailing
+# element — the latter is fragile if `all_manifests` is ever empty.
+name_args=()
+for m in "${all_manifests[@]}"; do
+    [[ ${#name_args[@]} -gt 0 ]] && name_args+=(-o)
```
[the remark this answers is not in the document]
devonjones: Won't fix — reclassifying P2 → P3 with justification per the Priority Mapping rule. The current line 85 doesn't contain an 'unset' or array-element removal at all; round 2 (commit 7e10667) replaced 'unset name_args[index]' with conditional '-o' insertion ([[ ${#name_args[@]} -gt 0 ]] && name_args+=(-o)). The ${#name_args[@]} reference that remains is a length check, not indexing. Your suggested array-slicing alternative addresses code that no longer exists. Misclassified — not a P2 correctness issue in the current code.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -0,0 +1,129 @@
...
+    if ! lookup="$(manifest_to_lang_priority "$mfile")"; then
+        continue
+    fi
+    lang="${lookup% *}"
+    priority="${lookup##* }"
+
+    entry="$(jq -cn \
+        --arg language "$lang" \
+        --arg subtree "$subtree" \
+        --arg manifest "$mfile" \
```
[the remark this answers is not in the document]
devonjones: Won't fix — reclassifying P2 → P3 with justification. Performance concern is real in principle, but bounded N: detect-language.sh runs ONCE per pr-review-loop invocation, only when the repo has no AGENT-REVIEWERS.md (i.e., first-time-setup path), and only iterates over found language-manifest files (typically 1-3 per repo, ~50 max in a giant Go workspace). The jq-per-iteration overhead is negligible at that scale; the refactor to a single jq call is purely stylistic at v1. Filed as P3 perf-polish; can revisit if profiling ever shows it as a hot path.

plugins/pr-review-loop/skills/pr-review-loop/scripts/install-template.sh
```diff
@@ -0,0 +1,298 @@
...
+fi
+
+# ---------------------------------------------------------------------------
+# Phase 2: extract pieces from each template and compute merged config.
+# ---------------------------------------------------------------------------
+
+# extract_section <template-path> <section-name>
+# Emits everything between `# <section-name>` and the next H1 (exclusive).
+# Accepts any whitespace between `#` and the heading text so templates
+# with `#  Agents` or `#\tAgents` parse correctly (not just `# Agents`).
```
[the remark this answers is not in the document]
devonjones: Won't fix — reclassifying P2 → P3 with justification. Round 2 already moved extract_section() from substr($0, 3) to a sub()-based approach. Your suggested regex /^# / matches ONLY '# Agents' (one space) and '#  Agents' (multiple spaces). My current regex /^#[[:space:]]+/ matches '# Agents', '#  Agents', AND '#\tAgents' (tab). Taking your suggestion would regress the tab case. The current code already uses the portable sub()-based approach you reference; the regex difference is intentional and strictly more permissive.

plugins/pr-review-loop/skills/pr-review-loop/scripts/install-template.sh
```diff
@@ -0,0 +1,298 @@
...
+        /^#[[:space:]]+/ {
+            if (in_section) { in_section = 0 }
+            heading = $0
+            sub(/^#[[:space:]]+/, "", heading)
+            sub(/[[:space:]]+$/, "", heading)
+            if (heading == section) { in_section = 1; next }
+        }
+        in_section { print }
+    ' "$file"
+}
```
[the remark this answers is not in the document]
devonjones: Won't fix — reclassifying HIGH → P3 with verification justification per the Priority Mapping rule. The regex /^#[[:space:]]+/ matches ONLY H1 (single '#' followed by whitespace), not H2. Direct verification:

  echo '# Agents'               | awk '/^#[[:space:]]+/ {print}'  →  MATCH
  echo '## test-coverage'       | awk '/^#[[:space:]]+/ {print}'  →  NO MATCH (second char is '#', not [[:space:]])
  printf '#\tAgents\n'        | awk '/^#[[:space:]]+/ {print}'  →  MATCH (tab)

End-to-end: installing the golang template extracts all 8 H2 agents under # Agents correctly (test I1 verifies this; 46/46 tests pass). The H2 'agent' headings do NOT prematurely close the # Agents section because /^#[[:space:]]+/ requires whitespace immediately after the FIRST '#'.

Your suggested /^#[[:space:]]+[^#]/ is more restrictive (rejects '# ## test'-style literal-multi-# heading text) but doesn't address the claimed bug because the claimed bug doesn't exist. HIGH-classification was based on a misread of the regex.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -0,0 +1,129 @@
...
+name_args=()
+for m in "${all_manifests[@]}"; do
+    [[ ${#name_args[@]} -gt 0 ]] && name_args+=(-o)
+    name_args+=(-name "$m")
+done
+
+# Run find inside the repo root so paths are clean relative.
+found_files="$(
+    cd "$REPO_ROOT"
+    find . "${prune_args[@]}" -type f \( "${name_args[@]}" \) -print 2>/dev/null \
```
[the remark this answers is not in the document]
devonjones: Fixed - added an early-exit branch (emit '[]' and exit 0) when name_args is empty. Today the manifest list is hardcoded with 4 entries so this can't happen; cheap to add now in case the list ever becomes config-driven and worth the defensive guard against a confusing find-syntax-error crash.

plugins/pr-review-loop/skills/pr-review-loop/scripts/install-template.sh
```diff
@@ -0,0 +1,298 @@
...
+    configs_array+="${sep}${cfg}"
+    sep=","
+done
+configs_array+="]"
+
+# Compute the merged config:
+#   - defaults_version_checked: current plugin version
+#   - disabled: union of all packs' disabled arrays (sorted, deduped)
+#   - overlap_acknowledged: union; conflict on same key with different value is fatal
+MERGED_CONFIG="$(printf '%s\n' "$configs_array" | jq --arg version "$CURRENT_PLUGIN_VERSION" '
```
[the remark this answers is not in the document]
devonjones: Won't fix — reclassifying P2 → P3 with verification justification. The script already has set -euo pipefail. With pipefail, the 'printf | jq' pipeline propagates jq's non-zero exit (jq error() exits with rc 5). Command substitution propagates non-zero through MERGED_CONFIG=$(...). Plus there's an explicit empty-check safety net ('if [[ -z "$MERGED_CONFIG" ]]; exit 1') as belt-and-suspenders. Tested: 'foo=$(jq -n "error(...)")' under set -e + pipefail exits non-zero as expected.

MAX_ARG_STRLEN: configs_array is small in practice (~1KB per pack; even 100 packs is ~100KB, well under the 128KB limit). The temp-file refactor would add complexity (file cleanup on exit) for no measurable benefit. Filed as a P3 perf-polish if profiling ever shows it as a hot path.

PR #27, opened 2026-05-20
feat(pr-review-loop): terraform + rails + javascript language templates + 1.5.0 bump (fv0)

devonjones:
## Summary

Closes the **fv0 ticket**: three new language-aware templates ship alongside the existing golang.md / python.md (from PR #25), plus a critical latent-bug fix found while drafting them.

Bumps pr-review-loop plugin **1.4.0 → 1.5.0** (feature add, SemVer minor).

### Three new templates (`plugins/pr-review-loop/agent-reviewer-templates/`)

| Template | Reviewers | Notes |
|----------|-----------|-------|
| `rails.md` | 8 | test-coverage, n+1, strong-params, migration-safety, callback-overuse, mass-assignment, route-bloat, clarity. Rails-specific source→test placement table. |
| `terraform.md` | 7 | provider-pinning, variable-typing, IAM-least-privilege, hardcoded-values, state-management, lifecycle, clarity. Vendor-agnostic — no organization-specific layering rules. |
| `javascript.md` | 7 | test-coverage (auto-detects test runner from package.json), async-handling, error-handling, dangerous-html, typescript-strictness (only fires on .ts/.tsx), `**js-vs-ts-policy-reviewer**`, clarity. |

### The `js-vs-ts-policy-reviewer` (user-requested)

Explicitly **does not** propose JS→TS migration. Reviews `.js` as JS, `.ts` as TS. Watches OTHER reviewers' findings for migration-pushing language ("should be TypeScript", "convert to .ts", etc.) and downgrades them to P3 with an out-of-scope rationale. Codifies "language-pluralistic codebase" as a first-class policy.

### detect-language.sh additions

- **terraform**: `main.tf` / `versions.tf` / `providers.tf` / `variables.tf` / `outputs.tf` (any one signals an active Terraform module)
- **javascript**: `package.json` (with or without tsconfig.json — single combined pack)
- **rails**: `Gemfile` ONLY when grep finds `gem "rails"` (suppresses Sinatra / pure-Ruby false-positives)

### Critical latent bug fix: fence-aware `extract_section`

While drafting rails.md (which uses `# BAD` / `# GOOD` as inline Ruby comments inside fenced code blocks), discovered that `install-template.sh`'s `extract_section` function naively treats those lines as H1 boundaries. **This bug shipped in PR #25 unnoticed** — python.md was installing with only 5 of 13 reviewers because of the same `# BAD` / `# GOOD` Python comments. Golang was unaffected (Go uses `//` for comments).

The dogfood on PR #25 didn't catch this because devon-claude-skills has its own AGENT-REVIEWERS.md, so install-template.sh never ran during dogfood. **First public install of python.md from 1.4.0 would have shipped broken.**

Fix: extract_section is now fence-aware. Tracks an `in_fence` flag toggled by lines matching `/^```/`. Inside a fence, the H1-detection branch is skipped. Two new regression tests (I_FA1 for python, I_FA2 for golang) lock this in.

## Plugin bump 1.4.0 → 1.5.0

- `plugins/pr-review-loop/.claude-plugin/plugin.json`
- `.claude-plugin/marketplace.json` (pr-review-loop entry)
- `AGENT-REVIEWERS.md` (upstream repo's own; `defaults_version_checked`)
- All three new templates pinned to 1.5.0 internally (install-template.sh overrides at write-time per PR #26 convention)

## Test plan

- [x] `tests/test-language-templates.sh` 60/60 (was 46; +14 for new detectors, per-template install counts, same-dir polyglot edge cases, and the fence-aware regression)
- [x] `tests/test-discover-agents.sh` 102/102
- [x] `tests/test-load-defaults.sh` 23/23
- [x] **Total: 185 assertions passing**
- [ ] Dogfood — run the loop on this PR

## Migration

Existing user repos pinned to 1.4.0 will see the stale-pin warning on next loop invocation. Migration: re-run with `--skip-stale-check` and bump pin to 1.5.0 manually (audit-agents q2h still not built).


plugins/pr-review-loop/agent-reviewer-templates/javascript.md
```diff
@@ -0,0 +1,464 @@
...
+
+---
+
+# Guidelines
+
+A reviewer pack for JavaScript and TypeScript projects. **One pack handles both**. The pack's reviewers branch on `tsconfig.json` presence and file extension (`.js` vs `.ts`) to apply rules appropriate to the source language.
+
+## Policy: No pushy JS → TS migration
+
+This pack explicitly **does NOT** propose converting `.js` files to TypeScript. JS-and-TS mixed codebases are valid; pure-JS codebases are valid; pure-TS codebases are valid. The pack reviews each file in the language it's written in. The `js-vs-ts-policy-reviewer` agent codifies this stance and pushes back if any other reviewer (or a human) attempts to flag a `.js` file simply for being JS.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


Verbose: three parallel 'X codebases are valid' clauses. → 'JS-only, TS-only, and mixed codebases are all valid.' (one clause, same meaning)

plugins/pr-review-loop/agent-reviewer-templates/javascript.md
```diff
@@ -0,0 +1,464 @@
...
+|----------|----------------|
+| `test-coverage-reviewer` | `**/*.{js,jsx,ts,tsx,mjs,cjs}` excluding `**/*.{test,spec}.*` and `__tests__/` |
+| `async-handling-reviewer` | `**/*.{js,jsx,ts,tsx,mjs,cjs}` |
+| `error-handling-reviewer` | `**/*.{js,jsx,ts,tsx,mjs,cjs}` |
+| `dangerous-html-reviewer` | `**/*.{jsx,tsx,js,ts}` (any file that may render or evaluate strings) |
+| `typescript-strictness-reviewer` | `**/*.{ts,tsx}` ONLY (no-op on JS) |
+| `js-vs-ts-policy-reviewer` | Activates when ANY review (human or agent) appears to push JS→TS migration |
+| `clarity-reviewer` | `*.md` (README, docs), JSDoc/TSDoc in `.{js,ts,jsx,tsx}` |
+
+Skip reviewers whose file scope doesn't match the PR diff. The `typescript-strictness-reviewer` is a no-op on PRs that touch only `.js` files; it does NOT flag the absence of TS.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


Redundant: second sentence repeats the table row directly above ('ONLY (no-op on JS)'). Delete: 'The `typescript-strictness-reviewer` is a no-op on PRs that touch only `.js` files; it does NOT flag the absence of TS.'

plugins/pr-review-loop/agent-reviewer-templates/rails.md
```diff
@@ -0,0 +1,499 @@
...
+
+---
+
+# Guidelines
+
+A reviewer pack for Ruby on Rails projects. Each section under `# Agents` below is an independent reviewer prompt with its own scope, set of patterns to flag, and disposition.
+
+## How the pack runs
+
+Each reviewer runs independently and reports findings without coordination. A reviewer's silence on something is not an endorsement — it just means that reviewer didn't see anything in its scope.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


Verbose filler: 'A reviewer's silence on something is not an endorsement — it just means that reviewer didn't see anything in its scope.' States the obvious; absent from the JS/TF templates. Delete the second sentence.

plugins/pr-review-loop/agent-reviewer-templates/rails.md
```diff
@@ -0,0 +1,499 @@
...
+      <%= post.author.name %>
+    <% end %>
+
+Issue: Iterating @posts and accessing post.author triggers one SELECT
+per post for the authors. With @posts.count > 1, this is a classic N+1.
+Suggested fix: In PostsController#index, change `@posts = Post.all` to
+`@posts = Post.includes(:author)`. Verify with bullet or rack-mini-profiler.
+```
+
+Structured findings are diff-able, easy to triage, and easy to deduplicate when multiple reviewers flag the same line.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


Verbose filler after the output example: 'Structured findings are diff-able, easy to triage, and easy to deduplicate when multiple reviewers flag the same line.' Not in other templates; the example speaks for itself. Delete.

plugins/pr-review-loop/agent-reviewer-templates/terraform.md
```diff
@@ -0,0 +1,436 @@
...
+  ],
+  "overlap_acknowledged": {}
+}
+```
+
+**Why most defaults are disabled:**
+
+Terraform is declarative IaC, not imperative code. Several baked-in default reviewers don't have a useful lens on `.tf` files:
+
+- **`silent-failure-hunter`** — Terraform doesn't have try/catch or error-swallowing patterns; resources either apply or they don't. Disabled.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


Redundant trailing word: '...resources either apply or they don't. Disabled.' The entry is already in the `disabled` list — the closing 'Disabled.' is noise. Delete it.

plugins/pr-review-loop/skills/pr-review-loop/tests/test-language-templates.sh
```diff
@@ -320,6 +320,174 @@ run_install "$repo" iT4 "golang:sub/../escape"
...
+echo "=== Test I_FA3: rails.md installs all 8 reviewers (fence-aware regression) ==="
+repo="$(make_temp_repo)"
+run_install "$repo" iFA3 rails:.
+agent_count="$(awk '
+    BEGIN { f=0; in_fence=0 }
+    /^```/ { in_fence = !in_fence; next }
+    !in_fence && /^# Agents/ { f=1; next }
+    !in_fence && /^# / { f=0 }
+    f && !in_fence && /^## / { print }
+' "$repo/AGENT-REVIEWERS.md" | wc -l)"
```
devonjones: 🤖 **Claude Code** (code-simplifier):


simplify: MEDIUM — Repeated awk-and-count pattern across 5 new tests (I_FA1–I_FA5)

The block from line ~389 to line ~404 (and four near-identical copies for terraform, javascript, python, golang) is the same 12-line shape every time: same awk program counting agents under '# Agents', same if/else, same PASS/FAIL/TESTS_PASSED/FAILED bookkeeping. The only things that vary per test are the repo path and the expected count.

This file already establishes a clear convention of small assert_* helpers (assert_jq, assert_exit, assert_file_exists, assert_file_grep at lines 53–116). The new tests should use the same pattern. Extracting a helper preserves behavior exactly and cuts ~60 lines down to ~5.

Suggested fix (add once near the other helpers, then call from each test):

```bash
assert_agent_count() {
    local name="$1" path="$2" expected="$3"
    local actual
    actual="$(awk '
        BEGIN { f=0; in_fence=0 }
        /^```/ { in_fence = !in_fence; next }
        !in_fence && /^# Agents/ { f=1; next }
        !in_fence && /^# / { f=0 }
        f && !in_fence && /^## / { print }
    ' "$path" | wc -l)"
    if [[ "$actual" == "$expected" ]]; then
        echo "  PASS  $name"; TESTS_PASSED=$((TESTS_PASSED + 1))
    else
        echo "  FAIL  $name (got $actual, expected $expected)"
        TESTS_FAILED=$((TESTS_FAILED + 1)); FAILED+=("$name")
    fi
}
```

Each I_FA test then collapses to:
```bash
run_install "$repo" iFA3 rails:.
assert_agent_count "rails.md installs all 8 reviewers" "$repo/AGENT-REVIEWERS.md" 8
```

Behavior identical (same awk, same comparison, same counters/arrays). Touch-it-you-own-it applies — these five tests are all newly added in this PR.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -118,6 +137,14 @@ while IFS= read -r relpath; do
     lang="${lookup% *}"
     priority="${lookup##* }"
 
+    # Rails-specific: a plain Gemfile could be Sinatra or pure-Ruby tooling.
+    # Only count it as a Rails marker if the file actually references rails.
+    if [[ "$mfile" == "Gemfile" ]]; then
+        if ! grep -qE "^[[:space:]]*gem[[:space:]]+['\"]rails['\"]" "$REPO_ROOT/$relpath" 2>/dev/null; then
```
devonjones: 🤖 **Claude Code** (silent-failure-hunter):


MEDIUM: `grep -qE ... 2>/dev/null` here collapses three distinct outcomes into one ("not Rails"): exit 0 (matched), exit 1 (no match), and exit 2 (error — file unreadable, permission denied, binary file). Because exit 2's stderr is silenced, a real Rails Gemfile that briefly becomes unreadable (race after `find`, ACL, root-only mount, encoding glitch) is silently dropped and the user sees "no language detected" with no clue why.

**What could be hidden:** I/O errors (permission denied, EACCES), file race after `find`, locale/encoding errors on Gemfile content, binary-content detection by grep.

**Suggested fix** — keep stderr visible and explicitly distinguish exit 2 from exit 1:

```bash
if [[ "$mfile" == "Gemfile" ]]; then
    grep -qE "^[[:space:]]*gem[[:space:]]+['\"]rails['\"]" "$REPO_ROOT/$relpath"
    rc=$?
    if [[ $rc -eq 1 ]]; then
        continue                       # legit "no rails gem" — skip silently
    elif [[ $rc -ne 0 ]]; then
        echo "Warning: could not read $relpath (grep exit $rc); skipping" >&2
        continue
    fi
fi
```

Minimal fix if you don't want to add the rc branch: just drop `2>/dev/null` so grep's own error message surfaces. The `set -e` short-circuit risk is already avoided by the leading `!`.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -55,18 +55,37 @@ REPO_ROOT="$(cd "$REPO_ROOT" && pwd)"
...
+        setup.py)          echo "python 1" ;;
+        requirements.txt)  echo "python 2" ;;
+        # Terraform: any of the 5 canonical files indicates an active module.
+        # main.tf is the most common entry point; the others rank lower for
+        # the "which manifest do we report" tiebreaker but all signal Terraform.
+        main.tf)           echo "terraform 0" ;;
+        versions.tf)       echo "terraform 1" ;;
+        providers.tf)      echo "terraform 2" ;;
+        variables.tf)      echo "terraform 3" ;;
+        outputs.tf)        echo "terraform 4" ;;
```
devonjones, line 75: 🤖 **Claude Code** (pr-test-analyzer):


test: MEDIUM — Behavior 'Terraform manifest priority tiebreaker (main.tf wins over versions.tf when both present in the same dir)' uncovered: criticality 6/10. This PR introduces priorities 0-4 across the five Terraform manifests (lines 71-75) specifically so unique_by([.language, .subtree]) at the bottom picks main.tf as the canonical entry. D9 tests main.tf alone, D10 tests versions.tf alone, D14 tests main.tf in a subdir — none put multiple Terraform files in the same directory to verify the priority sort actually works. Python's analogue is exercised by D4; Terraform's is not. A regression that swapped the priority numbers would silently change which manifest is reported. Suggested test 'detect_terraform_picks_main_tf_when_multiple_tf_files_coexist': create main.tf + versions.tf + providers.tf in the same dir, assert exactly one entry and .[0].manifest == 'main.tf'.

plugins/pr-review-loop/skills/pr-review-loop/scripts/install-template.sh
```diff
@@ -180,11 +180,26 @@ fi
...
     awk -v section="$section" '
-        BEGIN { in_section = 0 }
-        /^#[[:space:]]+/ {
+        BEGIN { in_section = 0; in_fence = 0 }
+        # Toggle fence state on any line that opens or closes a code block.
+        # Inside a fence, treat the fence line itself as content (print if
+        # currently in_section) and skip the H1-detection branch entirely.
+        /^```/ {
+            in_fence = !in_fence
+            if (in_section) print
```
devonjones, line 198: 🤖 **Claude Code** (pr-test-analyzer):


test: MEDIUM — Behavior 'fenced code blocks inside a target section are preserved verbatim in the extracted output' uncovered: criticality 5/10. The new fence-aware extract_section (lines 196-200) does `if (in_section) print` on the fence line itself so that ```...``` blocks are kept in the section body. The I_FA1-I_FA5 tests verify that the COUNT of '## ' agent headings is right after fence-awareness, but they do not assert that the actual fenced content (the body of the code block) round-trips into the installed file. If someone refactored to `next` without `print`, the agent count would still be 8/13/etc. but every code example would silently vanish from the installed template. Suggested test 'install_preserves_code_block_content_inside_extracted_section': install a known pack (e.g. python:.) and grep the installed AGENT-REVIEWERS.md for a distinctive substring from a code example inside the # Agents section (e.g. a Python comment string from one of the GOOD/BAD examples). Assert the match is present.

plugins/pr-review-loop/skills/pr-review-loop/tests/test-language-templates.sh
devonjones: 🤖 **Claude Code** (shell-script-reviewer):


Under `set -euo pipefail`, the pipeline `awk '...' "$repo/AGENT-REVIEWERS.md" | wc -l` will abort the test script if awk exits non-zero (e.g., if the file was never created because the install step failed), rather than recording a FAIL. This pattern repeats in all five I_FA\* agent-count blocks (lines 395, 416, 437, 459, 480).

Suggest appending `|| true` to suppress awk failures and allow the test to report a meaningful FAIL:
```bash
agent_count="$(awk '...' "$repo/AGENT-REVIEWERS.md" | wc -l || true)"
```
Or, alternatively, check `$iFA3_exit` before running the awk pipeline so a failed install produces a FAIL assert rather than an opaque script abort.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
devonjones: 🤖 **Claude Code** (pr-test-analyzer):


test: MEDIUM — Behavior 'Gemfile inside a pruned dir (vendor/, .git, etc.) does NOT trigger rails detection' uncovered: criticality 4/10. The new Gemfile entry (line 82) plus the content check (lines 142-146) make Gemfile a new manifest source. The skip_dirs list (line 91) includes 'vendor', and vendored Ruby gems often contain their own Gemfile referencing rails. D8 verifies node_modules pruning for go.mod but no equivalent test ensures a Gemfile inside vendor/ (or .git) is pruned. A subtle change to the prune expression would falsely flag every consumer of a rails gem as a rails repo themselves. Suggested test 'detect_skips_gemfile_in_vendor_dir': create vendor/some-gem/Gemfile with `gem "rails"` and assert the detect-language.sh output is [].

plugins/pr-review-loop/agent-reviewer-templates/rails.md
```diff
@@ -0,0 +1,499 @@
...
+      "overlaps_with": "pr-test-analyzer",
+      "reason": "Different lenses: test-coverage-reviewer enforces Rails-specific test placement rules (controller→test/controllers, model→test/models, etc.) and require-before-merge discipline. pr-test-analyzer scores behavioral gaps on a criticality 1-10 axis. Both contribute independent signal during review loops."
+    }
+  }
+}
+```
+
+**Why each default is disabled:**
+
+- **`silent-failure-hunter`** — covered by `n-plus-one-query-reviewer` (catches the Rails-flavor silent failure: missing eager-loading causing latency degradation that doesn't error) plus the Rails community's established `rescue => e` discouragement (handled in callback-overuse and strong-params reviewers).
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


docs: MEDIUM — Inaccurate coverage claim for `silent-failure-hunter`

The rationale states it is covered by `callback-overuse` and `strong-params` reviewers for the Rails community's `rescue => e` discouragement. Neither of those reviewers mentions rescue, exception swallowing, or error handling anywhere in their instructions — so the coverage claim is false.

The `n-plus-one-query-reviewer` part of the sentence is accurate (it catches the missing-eager-load silent failure). The second half is not supported by the pack's actual content.

Suggested fix: Remove the unsupported parenthetical, or point to the reviewer that actually covers it (none does currently, so the honest fix is to drop the claim):

```
- `silent-failure-hunter` — covered by `n-plus-one-query-reviewer` (catches the Rails-flavor silent failure: missing eager-loading causing latency degradation that doesn't error). `rescue => e` discouragement is a Rails community convention not separately enforced by this pack.
```

plugins/pr-review-loop/agent-reviewer-templates/javascript.md
```diff
@@ -0,0 +1,464 @@
...
+(e.g., `BD-XXX: JS→TS migration phase 2`) and the migration can be
+addressed as deliberate, scoped work — not as a per-PR nag.
+
+Downgrading this finding to P3 and recommending it be dropped from
+the review loop's exit-condition consideration.
+```
+
+**Implementation note:**
+
+This reviewer runs LAST in the round, after all other reviewers have posted. It scans the round's findings (via `get-review-comments.sh <PR>`) for migration-pushing language:
```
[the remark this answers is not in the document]
devonjones: Fixed - reframed the implementation note to acknowledge that the orchestrator spawns reviewers in parallel; the policy-reviewer now scans the PR's existing review comments (catching migration-pushing language across rounds), explicitly no longer requiring strict 'runs LAST' ordering. The previous wording overpromised what the orchestrator supports — good catch.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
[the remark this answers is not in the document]
devonjones: Fixed - applied the suggested regex (with the `\(?` paren tolerance) and verified `gem("rails")` now resolves to a rails detection. The same edit also distinguishes grep exit 1 (no match) from exit 2+ (I/O error) per silent-failure-hunter's parallel finding on this line.

plugins/pr-review-loop/agent-reviewer-templates/javascript.md
devonjones: Fixed - collapsed to 'JS-only, TS-only, and mixed codebases are all valid.'

plugins/pr-review-loop/agent-reviewer-templates/javascript.md
devonjones: Fixed - deleted the redundant sentence. The table row's '(no-op on JS)' already conveys it.

plugins/pr-review-loop/agent-reviewer-templates/rails.md
devonjones: Fixed - deleted the second sentence.

plugins/pr-review-loop/agent-reviewer-templates/rails.md
devonjones: Fixed - deleted the trailing line after the example.

plugins/pr-review-loop/agent-reviewer-templates/terraform.md
devonjones: Fixed - dropped the trailing 'Disabled.'

plugins/pr-review-loop/skills/pr-review-loop/tests/test-language-templates.sh
devonjones: Fixed - extracted `assert_agent_count` next to the other assert_* helpers; refactored I_FA1-I_FA5 to use it. ~60 lines → ~5 lines per test, behavior identical (same awk, same comparison, same TESTS_PASSED/FAILED bookkeeping). All 60 tests still pass locally. The helper also carries shell-script-reviewer's `|| true` defensiveness in one place rather than five.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
devonjones: Fixed - restructured to use an if/else around grep so set -e doesn't fire on exit 1, then $? captures the actual rc inside the else branch. Exit 1 → continue silently; exit 2+ → warn to stderr and continue. Dropped `2>/dev/null` so grep's own diagnostics surface. Combined with Gemini's parallel regex tightening on this same line.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
devonjones: Out of scope - tracked in devon-claude-skills-5fm. Criticality 6 (should-test, not blocking-gap per the pack's own scale); deferred to keep this PR focused on the templates + fence-aware fix. Ticket has the proposed test verbatim.

plugins/pr-review-loop/skills/pr-review-loop/scripts/install-template.sh
devonjones: Out of scope - tracked in devon-claude-skills-5fm. Criticality 5; deferred. Ticket includes the suggested 'install_preserves_code_block_content_inside_extracted_section' test.

plugins/pr-review-loop/skills/pr-review-loop/tests/test-language-templates.sh
devonjones: Fixed - the new `assert_agent_count` helper (next to the other assert_* helpers) wraps the awk pipeline with `|| true`, so a missing/malformed AGENT-REVIEWERS.md now returns count 0 and produces a FAIL line rather than aborting the suite under `set -euo pipefail`. Same defensiveness applied once instead of five times.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
devonjones: Out of scope - tracked in devon-claude-skills-5fm. Criticality 4; deferred. The vendored-Gemfile concern is real but D8 already exercises the prune mechanism for go.mod, so the regression risk is narrow.

plugins/pr-review-loop/agent-reviewer-templates/rails.md
devonjones: Fixed - dropped the unsupported parenthetical. Verified via grep across rails.md that neither callback-overuse-reviewer nor strong-params-reviewer mentions rescue, exception handling, or error swallowing. The honest replacement is your suggested wording: `rescue => e` discouragement is a community convention this pack doesn't separately enforce.

plugins/pr-review-loop/agent-reviewer-templates/javascript.md
```diff
@@ -0,0 +1,466 @@
...
+(e.g., `BD-XXX: JS→TS migration phase 2`) and the migration can be
+addressed as deliberate, scoped work — not as a per-PR nag.
+
+Downgrading this finding to P3 and recommending it be dropped from
+the review loop's exit-condition consideration.
+```
+
+**Implementation note:**
+
+This reviewer scans the PR's existing review comments (via `get-review-comments.sh <PR>`) for migration-pushing language. It runs in parallel with the other reviewers; in any given round it will only see findings that other reviewers (or humans) posted in earlier rounds. That's fine — across the lifetime of a PR's review loop, migration-pushing language gets caught and replied to within one or two rounds of being posted. The reviewer does NOT require strict "runs last" ordering, which the orchestrator does not currently support.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


Verbose: the new implementation note packs three ideas into one run-on sentence. 'That's fine — across the lifetime of a PR's review loop, migration-pushing language gets caught and replied to within one or two rounds of being posted.' is editorial reassurance that adds length without precision; cut it. 'which the orchestrator does not currently support' is an aside; fold the constraint into the main clause instead. Suggested rewrite: 'This reviewer scans existing PR review comments (via `get-review-comments.sh <PR>`) for migration-pushing language. It runs in parallel with other reviewers and only sees findings from earlier rounds; strict last-ordering is not required.'

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -118,6 +137,25 @@ while IFS= read -r relpath; do
     lang="${lookup% *}"
     priority="${lookup##* }"
 
+    # Rails-specific: a plain Gemfile could be Sinatra or pure-Ruby tooling.
+    # Only count it as a Rails marker if the file actually references rails.
+    # Distinguish grep exit 1 (no match — not Rails, skip silently) from exit 2+
+    # (I/O error — stderr stays visible, warn and skip).
+    # Regex accepts both `gem "rails"` and `gem("rails")` syntax.
+    if [[ "$mfile" == "Gemfile" ]]; then
+        if grep -qE "^[[:space:]]*gem[[:space:]]*\(?[[:space:]]*['\"]rails['\"]" "$REPO_ROOT/$relpath"; then
```
devonjones: 🤖 **Claude Code** (pr-test-analyzer):


test: MEDIUM — Behavior 'widened Gemfile regex accepts gem('rails') paren syntax and single-quoted gem 'rails'' uncovered: criticality 5/10. Round-1 fix (804beb6) deliberately widened the regex from `gem[[:space:]]+['"]rails['"]` to `gem[[:space:]]*\(?[[:space:]]*['"]rails['"]` specifically to accept both `gem("rails")` and `gem 'rails'` in addition to the existing `gem "rails"`. D11 (line 368-371) only exercises double-quoted-no-parens; the two NEW syntactic forms the commit added support for have zero test coverage. A regression that drops `\\(?` or narrows the quote class would silently fail to detect Rails repos using either non-default style — exactly the false-negative the fix was meant to prevent. This is distinct from 5fm's T3 (vendor-pruning) and a new gap introduced by THIS commit, not a pre-existing one. Suggested test 'detect_rails_via_gemfile_with_paren_and_single_quote_syntax': two assertions in a single test — (a) Gemfile containing `gem('rails', '~> 7.1')` → detects rails; (b) Gemfile containing `gem 'rails'` (single-quoted, no parens) → detects rails. Together they pin both new branches of the widened regex.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -118,6 +137,25 @@ while IFS= read -r relpath; do
     lang="${lookup% *}"
     priority="${lookup##* }"
 
+    # Rails-specific: a plain Gemfile could be Sinatra or pure-Ruby tooling.
+    # Only count it as a Rails marker if the file actually references rails.
+    # Distinguish grep exit 1 (no match — not Rails, skip silently) from exit 2+
+    # (I/O error — stderr stays visible, warn and skip).
+    # Regex accepts both `gem "rails"` and `gem("rails")` syntax.
+    if [[ "$mfile" == "Gemfile" ]]; then
```
devonjones, line 145: 🤖 **Claude Code** (code-simplifier):


simplify: LOW — Gemfile rails-content check is wordier than it needs to be

The author's commit note pre-empts a flag here by saying this matches the existing style in this script. Re-reading the script though, that's not quite right — the only other error-handling site is line 134's `if ! lookup="$(manifest_to_lang_priority "$mfile")"; then continue; fi` (a one-liner), and the *previous* incarnation of this very block was also a one-liner: `if ! grep -qE ... 2>/dev/null; then continue; fi`. There is no precedent in this script for the new four-branch if/else with the `:  # matched` no-op limb. Both `cmd || rc=$?` and the new if/else are equally new patterns; one is just shorter.

The `cmd || rc=$?` capture is the canonical set-e-safe exit-code idiom and collapses 13 lines (145-157) to 5 with identical behavior:

```bash
if [[ "$mfile" == "Gemfile" ]]; then
    rc=0
    grep -qE "^[[:space:]]*gem[[:space:]]*\(?[[:space:]]*['\"]rails['\"]" "$REPO_ROOT/$relpath" || rc=$?
    if [[ $rc -ne 0 ]]; then
        [[ $rc -ne 1 ]] && echo "Warning: could not read $relpath (grep exit $rc); skipping" >&2
        continue
    fi
fi
```

Behavior identical: exit 0 → fall through to entry emission; exit 1 → silent continue; exit ≥2 → warn to stderr then continue. `rc=0` initialization guards against stale values. Linter-friendly (no `:` no-op limb).

If you prefer the explicit ladder, that's reasonable too — the no-op `:` limb is the bit that reads awkwardly because it inverts the natural success-falls-through flow. Either of these reads more naturally:

```bash
# Variant: keep the ladder, drop the no-op limb
rc=0
grep -qE "..." "$REPO_ROOT/$relpath" || rc=$?
case $rc in
    0) ;;                                  # matched, proceed
    1) continue ;;                         # no match, skip silently
    *) echo "Warning: ... (grep exit $rc); skipping" >&2; continue ;;
esac
```

Behavior-preserving in all cases. Touch-it-you-own-it: the whole block is new in commit 804beb6.

plugins/pr-review-loop/agent-reviewer-templates/javascript.md
devonjones: Fixed - applied your suggested rewrite verbatim. The original reframing did indeed pack three ideas into one run-on sentence; the editorial 'That's fine' and the trailing aside were both fat. New text is two clean sentences.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
devonjones: Out of scope - added to devon-claude-skills-5fm as T4. Criticality 5 (should-test, not blocking-gap per the pack's own scale). The beads ticket now carries the verbatim two-assertion test ('detect_rails_via_gemfile_with_paren_and_single_quote_syntax') alongside T1/T2/T3.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
devonjones: Fixed - you're right, my 'matches existing style' pre-emption was wrong (the prior code at this site was the one-liner you cited). Applied your case-statement variant; the 3-state flow reads naturally (0 falls through, 1 continues silently, * warns and continues). 13 lines → 7. 60/60 tests still pass.

plugins/pr-review-loop/agent-reviewer-templates/terraform.md
```diff
@@ -0,0 +1,436 @@
...
+
+## iam-least-privilege-reviewer
+
+Review IAM policies (AWS `aws_iam_policy_document`, `aws_iam_role_policy`, GCP `google_*_iam_binding`, Azure `azurerm_role_assignment`, etc.) for **principle of least privilege**.
+
+**Patterns to flag:**
+
+### `Action: "*"` + `Resource: "*"` on a non-admin role → P1
+
+This grants full cloud admin. Only flag-not-flag for genuine admin roles (`administrator`, `breakglass`, etc.). Otherwise flag.
```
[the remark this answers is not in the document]
devonjones: Fixed - applied your suggested wording. 'Only exempt genuine admin roles' resolves the ambiguous 'flag-not-flag' phrasing cleanly.

plugins/pr-review-loop/skills/pr-review-loop/tests/test-language-templates.sh
```diff
@@ -115,6 +115,29 @@ assert_file_not_grep() {
...
+assert_agent_count() {
+    local name="$1" path="$2" expected="$3"
+    local actual
+    actual="$(awk '
+        BEGIN { f=0; in_fence=0 }
+        /^```/ { in_fence = !in_fence; next }
+        !in_fence && /^# Agents/ { f=1; next }
+        !in_fence && /^# / { f=0 }
+        f && !in_fence && /^## / { print }
+    ' "$path" 2>/dev/null | wc -l || true)"
```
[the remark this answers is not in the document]
devonjones: Fixed - rewrote assert_agent_count synthesizing all three of your concerns:

1. **wc -l portability**: counts inside awk's END block (your suggestion). No `wc -l`, no BSD/macOS leading-space gotcha.
2. **Flexible H1 matching**: `^#[[:space:]]+` + `sub()` to extract heading text, then exact `h=="Agents"` match — mirrors install-template.sh's own parser. Verified by sanity test: a file with `#\tAgents` (tab between hash and 'Agents') now counts correctly.
3. **`|| true` masking errors**: removed. Replaced with an upfront file-exists guard that records a clear FAIL line for missing files. This preserves shell-script-reviewer's round-1 concern (don't abort suite on missing file) without the false-PASS-at-expected=0 risk you identified.

60/60 tests still pass.

plugins/pr-review-loop/agent-reviewer-templates/javascript.md
```diff
@@ -0,0 +1,466 @@
...
+
+Migration-pushing language to scan for:
+
+- "should be TypeScript"
+- "convert to .ts"
+- "use TypeScript instead"
+- "add types" *in the context of a .js file*
+- "rename .js to .ts"
+
+When such a suggestion is found, post the policy comment as a reply (resolving the migration suggestion as P3 / out-of-scope).
```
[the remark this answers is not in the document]
devonjones: Fixed - same family of critique as your round-1 G1 (which caught the 'runs LAST' ordering gap); this one caught the action-type gap. Both addressed in commit 41f28bd:

1. **`<PR>` placeholder**: replaced the explicit `get-review-comments.sh <PR>` with 'the orchestrator supplies the PR number and review-fetching tooling at spawn time' — the literal placeholder inside agent instructions isn't substituted by the spawn template, so describing the contract abstractly is more honest.

2. **Reply/resolve actions**: reframed to 'emit the policy comment as a standard agent-reviewer finding (e.g., via `post-line-comment.sh`); the orchestrator's main loop handles linking back to the source thread and dispositioning it as P3 / out-of-scope.' Agent reviewers' actual contract is to emit findings, not manage threads — the new wording matches.

plugins/pr-review-loop/skills/pr-review-loop/scripts/detect-language.sh
```diff
@@ -118,6 +137,21 @@ while IFS= read -r relpath; do
...
     priority="${lookup##* }"
 
+    # Rails-specific: a plain Gemfile could be Sinatra or pure-Ruby tooling.
+    # Only count it as a Rails marker if the file actually references rails.
+    # Regex accepts `gem "rails"`, `gem 'rails'`, and `gem("rails")` syntax.
+    # Distinguish grep exit 1 (no match — skip silently) from exit 2+ (I/O error
+    # — warn to stderr; grep's own stderr also visible since no `2>/dev/null`).
+    if [[ "$mfile" == "Gemfile" ]]; then
+        rc=0
+        grep -qE "^[[:space:]]*gem[[:space:]]*\(?[[:space:]]*['\"]rails['\"]" "$REPO_ROOT/$relpath" || rc=$?
```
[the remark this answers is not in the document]
devonjones: Reclassifying P2 → P3 (style/preference, not correctness) per the Priority Mapping rule, and declining.

The suggested regex doesn't actually solve the stated concern. The proposed pattern `(\(?[[:space:]]*['"]rails['"]([[:space:]]*,|\)?))` ends with `([[:space:]]*,|\)?)` — the alternative branch `\)?` matches **zero or one** closing paren, so a malformed `gem("rails"` (missing close paren) would still match. The regex doesn't enforce balanced parens.

For real-world Gemfile content the current regex `^[[:space:]]*gem[[:space:]]*\(?[[:space:]]*['"]rails['"]` already accepts the three valid syntactic forms we widened to cover (gem "rails", gem 'rails', gem("rails")). I cannot construct a real Gemfile line that the suggested regex would reject and the current regex would falsely accept — both behave identically for valid inputs and similarly for malformed ones.

The current pattern reads as a heuristic (which it is). Adding the trailing alternation group adds complexity without measurable precision gain. Keeping the simpler form.

Not a self-contradiction with round-1 G2 (which deliberately widened this same regex). G2 added syntactic coverage; G6 would add structural validation that the suggested pattern doesn't actually deliver.

PR #31, opened 2026-05-28
fix(pr-review-loop): sweep fence-unaware awk bug class to 2 more scripts

devonjones:
## Summary

A downstream review surfaced that the earlier `extract_section` fence-aware fix in `install-template.sh` (#27) only closed one of three occurrences of the same silent-data-loss failure mode. This PR sweeps the remaining two — both silent failures that would produce mysterious user-facing symptoms ("loop randomly loses reviewers", "disable list ignored").

### `scripts/discover-agents.sh` (P1 — runtime path)

The awk in `parse_agent_file` had no `in_fence` state. Worst case: any agent whose body contained a Python/Ruby/shell fenced example with `# BAD` or `## Example` lines would silently `flush()` mid-agent, flip `in_agents` off (because the bogus h1 wasn't "Agents"), and **drop every agent defined below that point** from the JSON output.

More severe than the install-template instance because it's the *runtime* agent-spawn path — silent reviewer loss on every loop invocation with no error surface.

### `scripts/_parse_configuration.sh` (P1 — config path)

The `in_fence` variable here tracked only the json fence, not generic code fences. The unconditional `/^# / { in_config = 0 }` rule fired on any `^# ` line. Today the canonical template hides the bug, but the parser contract permits prose before the fence — and an explanatory fenced shell example with `#` comments would silently flip `in_config` off before the json block was reached. **User's `disabled` / `overlap_acknowledged` / `independent_validator` settings silently discarded** with no error surface.

### `scripts/discover-agents.sh` line ~283 (cosmetic)

The `grep -q '^# Configuration[[:space:]]*$'` check that emits the "subdirectory # Configuration ignored" warning was also fence-unaware — would false-positive on fenced examples in subtree agent bodies. Replaced with a small fence-aware awk. No data loss, just noise.

## Approach

Ported the `in_fence` toggle pattern from `install-template.sh:189-209` (PR #27). The toggle fires on any line starting with three backticks; H1/H2 detection is guarded by `!in_fence`. For `_parse_configuration.sh`, separated generic fence tracking from the json-block-capture state so a single fence pattern covers both the bogus-heading case and the json-block boundary.

## Tests

Generalized the I_FA1-5 regression pattern (PR #27) into four new tests in `test-discover-agents.sh`:

- **Test 29:** fenced `# ` inside agent body doesn't drop subsequent agents (3 agents → 3 agents)
- **Test 30:** fenced `## ` inside agent body doesn't fragment current agent or create spurious siblings
- **Test 31:** fenced `^# bash example` in narrative prose before the json block doesn't discard `# Configuration`
- **Test 32:** fenced `# Configuration` inside subtree agent body doesn't trigger the "ignored" warning

All three suites green: **117** (discover-agents) + **60** (language-templates) + **23** (load-defaults) = **200/200**.

## Test plan

- [x] `tests/test-discover-agents.sh` (117/117)
- [x] `tests/test-language-templates.sh` (60/60)
- [x] `tests/test-load-defaults.sh` (23/23)
- [ ] Dogfood pr-review-loop on this PR


plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -173,9 +173,23 @@ parse_agent_file() {
...
+        in_fence = 0
     }
 
-    /^# / {
+    # Toggle fence state on any line starting with three backticks. Inside a
+    # fence, treat the line as content (so the rendered agent body keeps its
+    # opening/closing fence) and skip the H1/H2-detection branches. Without
+    # this, an agent body containing a fenced `# BAD` or `## Example` line
+    # silently flushes mid-agent and (worse) flips `in_agents` off when the
+    # bogus h1 does not equal "Agents", dropping every agent defined below.
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


Comment conflates two different failure modes under one description. The wording 'an agent body containing a fenced `# BAD` or `## Example` line silently flushes mid-agent and (worse) flips `in_agents` off when the bogus h1 does not equal "Agents"' implies both cases hit the same code path. They don't: `# BAD` triggers the H1 rule (which sets in_agents=0 since 'BAD'!='Agents', dropping subsequent agents — this is the '(worse)' case); `## Example` triggers the H2 rule (only when in_agents=1), creating a spurious agent named 'Example' but leaving in_agents intact. Test 29 exercises the H1 path; Test 30 exercises the H2 path and asserts no spurious 'Installation'/'Usage' agents — two distinct symptoms. A reader matching this comment to Test 30's assertions will be confused about what fails. Split the comment into the two cases, e.g. 'a fenced `# BAD` line silently flushes the current agent AND flips in_agents off (dropping every agent below); a fenced `## Example` line silently creates a spurious agent named Example.'

plugins/pr-review-loop/skills/pr-review-loop/tests/test-discover-agents.sh
```diff
@@ -707,6 +707,158 @@ run_discover "$repo" "sub/file.go" t28
...
+    }
+  }
+}
+```
+EOF
+run_discover "$repo" "src/foo.py" t31
+assert_exit "exit 0" "$t31_exit" "0"
+assert_jq "disabled list parsed from json (would be empty under fence-unaware parser)" "$t31" '.configuration.disabled_defaults == ["code-simplifier"]'
+assert_jq "overlap_acknowledged parsed" "$t31" '.configuration.overlaps_acknowledged | has("my-extra-reviewer")'
+rm -rf "$repo"
```
devonjones, line 817: 🤖 **Claude Code** (pr-test-analyzer):


Coverage gap (criticality 7): no test for **multiple ```json blocks under `# Configuration`**. The awk in `_parse_configuration.sh` (lines 50-77) `exit`s as soon as the first json fence closes, so if a user writes a documentation ```json example block followed by the real ```json config block under the same `# Configuration` heading, only the EXAMPLE is parsed and the real config is silently ignored. Test 31 sidesteps this by using a ```bash example fence — please add a test where BOTH fences are ```json (e.g., an annotated 'default values look like this' example followed by the actual config) and assert which one wins (or that the parser errors / warns). This is a realistic doc pattern that will silently misconfigure the loop.

plugins/pr-review-loop/skills/pr-review-loop/tests/test-discover-agents.sh
```diff
@@ -707,6 +707,158 @@ run_discover "$repo" "sub/file.go" t28
...
+
+A third reviewer that should also still appear.
+EOF
+run_discover "$repo" "src/foo.py" t29
+assert_exit "exit 0" "$t29_exit" "0"
+assert_jq "first user agent present" "$t29" 'any(.agents[]; .name == "first-reviewer" and .kind == "user")'
+assert_jq "second user agent present (would be dropped by fence-unaware parser)" "$t29" 'any(.agents[]; .name == "second-reviewer" and .kind == "user")'
+assert_jq "third user agent present" "$t29" 'any(.agents[]; .name == "third-reviewer" and .kind == "user")'
+assert_jq "user_count is 3" "$t29" '.configuration.user_count == 3'
+rm -rf "$repo"
```
devonjones, line 747: 🤖 **Claude Code** (pr-test-analyzer):


Coverage gap (criticality 7): no test for **unclosed fence inside an agent body**. `parse_agent_file`'s awk (discover-agents.sh lines 185-190) toggles `in_fence` on every `^\`\`\`` but never resets it on H1/H2 or EOF. If an agent body opens a ```bash fence and forgets the closing ```, `in_fence` stays 1 for the rest of the file — every subsequent `## next-agent` heading falls into the catch-all content branch and gets silently swallowed into the prior agent's body. Same silent-data-loss failure mode the fence-aware fix was meant to close, just triggered by malformed input instead of by H1/H2 inside fences. Please add a test: agent A with an unclosed ``` followed by agent B, and assert that EITHER agent B still parses (preferred — auto-close at H1/H2/EOF) OR the parser exits non-zero with a clear 'unclosed fence' diagnostic. Currently it silently drops agent B.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -40,18 +40,40 @@ fi
...
-    /^# / { in_config = 0 }
-    in_config && /^```json[[:space:]]*$/ { in_fence = 1; next }
-    in_config && in_fence && /^```[[:space:]]*$/ { in_fence = 0; in_config = 0; exit }
-    in_config && in_fence { print }
+    BEGIN { in_config = 0; in_fence = 0; in_json = 0 }
+    # Track every fence open/close so heading detection stays accurate even
+    # when prose preceding the json block contains fenced `^# ` examples.
+    /^```/ {
+        # Json-block bookkeeping: if we are about to close the json fence we
+        # were capturing, emit nothing and exit cleanly.
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


Verbose: 'Json-block bookkeeping: if we are about to close the json fence we were capturing, emit nothing and exit cleanly.' paraphrases what the next 5 lines of code already show. Trim to one line, e.g. '# Close the captured json fence.' The exit/reset behavior is visible in the block.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -40,18 +40,40 @@ fi
...
+        if (in_config && !in_fence && /^```json[[:space:]]*$/) {
+            in_fence = 1
+            in_json = 1
+            next
+        }
+        # Any other fence: just toggle the generic tracker.
+        in_fence = !in_fence
+        next
+    }
+    !in_fence && /^# Configuration[[:space:]]*$/ { in_config = 1; next }
```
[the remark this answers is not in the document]
devonjones: Fixed - swept `[[:space:]]+` after `#` across all new H1 patterns in _parse_configuration.sh (Configuration match + heading-end-guard) and the new fence-aware subtree warning awk in discover-agents.sh. Pre-existing `/^# /` / `/^## /` in parse_agent_file (lines 192/203) left alone — out of scope for this fence-aware bug-class sweep; can flex those in a follow-up.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
devonjones: Fixed - split the conflated comment into two distinct bullets so a reader matches each test (H1 path = Test 29; H2 path = Test 30) to its specific failure mode.

plugins/pr-review-loop/skills/pr-review-loop/tests/test-discover-agents.sh
devonjones: Fixed - added Test 33 documenting first-wins semantics when multiple json blocks appear under `# Configuration`. Asserts the example block's `disabled` entry is parsed and the real block's entry is not — locking in current contract so a switch to last-wins or merge-all is caught explicitly.

plugins/pr-review-loop/skills/pr-review-loop/tests/test-discover-agents.sh
devonjones: Out of scope - tracked in devon-claude-skills-ar9. This PR closes one fence-unaware bug class (H1/H2 inside closed fences). Unclosed-fence handling is a different class (malformed-input recovery) that warrants its own design decision (warn-at-EOF vs auto-close-at-H1 vs hard-fail). Ticket captures the option space and recommends warn-at-EOF as least disruptive.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - trimmed to "# Close the captured json fence." — the exit/reset behavior is visible in the next 4 lines.

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -173,9 +173,26 @@ parse_agent_file() {
...
+        in_fence = !in_fence
+        if (content != "") content = content "\n" $0
+        else content = $0
+        next
+    }
+
+    !in_fence && /^# / {
         flush()
         h1_name = substr($0, 3)
         gsub(/^[[:space:]]+|[[:space:]]+$/, "", h1_name)
```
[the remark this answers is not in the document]
devonjones: Fixed - swept parse_agent_file H1 detection from `substr($0, 3)` + `/^# /` to flexible `sub(/^#[[:space:]]+/, ...)` + `/^#[[:space:]]+/`. Scope reconsideration from round 1: bundling here keeps the fence-aware sweep cohesive (one PR closes the bug class + brings the regex consistency).

plugins/pr-review-loop/skills/pr-review-loop/scripts/discover-agents.sh
```diff
@@ -186,7 +203,7 @@ parse_agent_file() {
         next
     }
 
-    /^## / && in_agents {
+    !in_fence && /^## / && in_agents {
         flush()
         current_h2 = substr($0, 4)
         gsub(/^[[:space:]]+|[[:space:]]+$/, "", current_h2)
```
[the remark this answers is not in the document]
devonjones: Fixed - same flexible-regex sweep applied to H2 detection in parse_agent_file: `substr($0, 4)` + `/^## /` → `sub(/^##[[:space:]]+/, ...)` + `/^##[[:space:]]+/`.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -40,18 +40,39 @@ fi
...
+        }
+        # Any other fence: just toggle the generic tracker.
+        in_fence = !in_fence
+        next
+    }
+    !in_fence && /^#[[:space:]]+Configuration[[:space:]]*$/ { in_config = 1; next }
+    # Any other H1 (outside fences) ends the Configuration section.
+    !in_fence && /^#[[:space:]]+/ { in_config = 0 }
+    in_json { print }
 ' "$FILE")"
```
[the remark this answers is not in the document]
devonjones: Fixed - END block added in both _parse_configuration.sh AND discover-agents.sh parse_agent_file (the two awk programs in this PR's fence-aware sweep). Tests 34/35 added to assert the stderr warning fires on malformed input. Closing devon-claude-skills-ar9 since the recommended option-1 (warn-at-EOF) is now landed in both files.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
```diff
@@ -40,18 +40,47 @@ fi
...
+        next
+    }
+    !in_fence && /^#[[:space:]]+Configuration[[:space:]]*$/ { in_config = 1; next }
+    # Any other H1 (outside fences) ends the Configuration section.
+    !in_fence && /^#[[:space:]]+/ { in_config = 0 }
+    in_json { print }
+    END {
+        # Defensive: warn on malformed input. An unclosed fence here means
+        # the json block was never closed cleanly, so RAW_JSON is incomplete
+        # and the downstream jq parse will fail with a less obvious error.
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


The END-block prose overstates one failure path. 'An unclosed fence here means the json block was never closed cleanly, so RAW_JSON is incomplete and the downstream jq parse will fail' — but in_fence=1 at EOF covers TWO cases: (a) the json fence opened and never closed (in_json=1, RAW_JSON incomplete, jq fails) and (b) a non-json prose fence opened and never closed (in_json=0, RAW_JSON empty, the [[ -z RAW_JSON ]] branch returns {} silently, jq never runs). Test 35 actually exercises case (b) — the outer ```bash never closes, so RAW_JSON stays empty and the fallback fires; the user sees the warning but no jq error. A reader matching the comment to Test 35 will be confused about why no jq parse error appears. Suggest broadening to: 'An unclosed fence leaves RAW_JSON either empty (non-json fence never closed → silent {} fallback) or incomplete (json fence never closed → cryptic jq error); warn in both cases.'

plugins/pr-review-loop/skills/pr-review-loop/tests/test-discover-agents.sh
```diff
@@ -707,6 +707,250 @@ run_discover "$repo" "sub/file.go" t28
...
+{
+  "disabled": ["code-simplifier"]
+}
+```
+EOF
+run_discover "$repo" "src/foo.py" t35
+# The unclosed fence corrupts the parse — but the script falls back to {}
+# rather than aborting the whole loop. What we care about is the user gets
+# a clear stderr signal so they can fix their markdown.
+assert_stderr_contains "unclosed-fence warning in stderr" "$t35_err" "unclosed code fence"
```
devonjones, line 870: 🤖 **Claude Code** (pr-test-analyzer):


Coverage gap (criticality 7): Test 35 asserts the stderr warning but never asserts the script's exit code. Test 34 (its sibling) uses `assert_exit "exit 0 (warning, not error)" "$t34_exit" "0"`. The doc-comment above the test explicitly states 'the script falls back to {} rather than aborting the whole loop' — that's the contract — but only the warning half of the contract is tested. If a future change makes unclosed-fence fatal in _parse_configuration.sh (e.g., someone adds `exit 1` next to the warn), the test silently passes as long as awk's END block prints the warning before the bash-level exit. Suggest adding `assert_exit "exit 0 (warning, not error)" "$t35_exit" "0"` to lock the non-fatal half of the contract. Optionally also assert `.configuration.disabled_defaults == []` against t35 stdout to lock the {} fallback explicitly.

plugins/pr-review-loop/skills/pr-review-loop/scripts/_parse_configuration.sh
devonjones: Fixed - expanded the END-block prose to call out both failure modes covered by `in_fence == 1` at EOF (case a: json fence unclosed → jq error; case b: prose fence unclosed → silent `{}` fallback). Case (b) is the actual silent-config-loss mode the warning surfaces — matches Test 35's scenario.

plugins/pr-review-loop/skills/pr-review-loop/tests/test-discover-agents.sh
devonjones: Fixed - added `assert_exit "$t35_exit" "0"` to Test 35 matching Test 34's pattern, locking in the non-fatal half of the contract.

PR #33, opened 2026-06-21
Add dream plugin + pr-review-loop marker integration

devonjones:
Reframes the standalone `dream` tool as a marketplace plugin with two skills, and wires pr-review-loop to feed it.

## `dream` plugin (two skills)
- **`dream`** — session miner: distill (heuristic friction + local-ollama enrichment, cached digests) → pool → Opus consolidation (dedup vs known corpus, frequency-gated promotion, behavior-vs-product-decision routing).
- **`dream-reviewers`** — validate the reviewer roster from PR history → per-reviewer scorecards (value + fix-strict acceptance, FP / won't-fix / reopen / taste-override) → keep/refine/retire proposals.

Fully generalized (derives repo, log dir, `~/.dream/<slug>` home, known-corpus from context). Derived data lives outside the repo; copy-on-write promotion.

## pr-review-loop integration (checkpoint-writer pattern)
pr-review-loop now emits a durable dream-marker per finding+disposition via `emit-dream-marker.sh` (best-effort), recording `disposition_by` so dream finally gets the operator-taste signal. Markers are dream's authoritative source; GitHub + log harvest are backfill. Contract: `plugins/dream/references/MARKER-CONTRACT.md`. Bumps pr-review-loop 1.5.0 → 1.6.0.


plugins/dream/hooks/session-start-proposals.sh
```diff
@@ -0,0 +1,53 @@
...
+# project's .claude/hooks/ and register it as a SessionStart "command" hook.
+#
+# Self-contained: it derives the SAME per-project slug the dream tool uses
+# (git-root basename — see dreamlib/config.project_slug) so it points at the same
+# ~/.dream/<slug>/ home, with no dependency on the dream scripts being importable.
+#
+# Contract: always exits 0 and never errors out — a SessionStart hook must never
+# block a session. Plain stdout is added to the session context, so the message
+# is phrased for the assistant to relay to the user.
+set -u
```
devonjones, line 15: 🤖 **Claude Code** (shell-script-reviewer):


New script uses `set -u` but is missing `-e` and `-o pipefail`. Change to `set -euo pipefail`. The no-error contract (lines 12-14) is already enforced by the `|| exit 0` / `|| true` guards throughout the script body, so adding `-e` and `pipefail` is safe and brings this in line with the repo standard.

plugins/pr-review-loop/skills/pr-review-loop/scripts/emit-dream-marker.sh
```diff
@@ -0,0 +1,47 @@
...
+# e.g.
+#   emit-dream-marker.sh reviewer-finding pr=702 round=1 \
+#       reviewer=complexity-reviewer severity=P3 file=x.py line=42 \
+#       disposition=wont_fix disposition_by=user finding="…" reason="…" validator=valid
+#
+# BEST-EFFORT: this must never fail or block the review loop. Any error (no jq,
+# not a git repo, write failure) silently no-ops. Records append to
+# ~/.dream/<git-root-basename>/markers/pr-review-loop.jsonl (or $DREAM_HOME).
+# See the dream plugin's references/MARKER-CONTRACT.md for the schema.
+
```
devonjones, line 15: 🤖 **Claude Code** (shell-script-reviewer):


New script has no `set -euo pipefail`. Even though the outer `_emit "$@" 2>/dev/null || true; exit 0` pattern makes this best-effort, the repo standard is `set -euo pipefail` in every script. Add it immediately after the header comment block (before `_emit()`); the existing `2>/dev/null || return 0` / `|| true` guards inside `_emit` already handle all the intentional swallowed-error paths.

plugins/dream/scripts/dreamlib/synth.py
```diff
@@ -0,0 +1,218 @@
...
+    for f in KNOWN_CORPUS_FILES:
+        if os.path.exists(f):
+            sources.append((os.path.basename(f), f))
+    for f in sorted(glob.glob(os.path.join(MEMORY_DIR, "*.md"))):
+        sources.append((f"memory/{os.path.basename(f)}", f))
+    for label, path in sources:
+        try:
+            with open(path) as fh:
+                text = fh.read()
+        except Exception:
```
devonjones, line 129: 🤖 **Claude Code** (silent-failure-hunter):


MEDIUM — silent corpus-read failure causes false 'novel' recommendations. `_load_known_paragraphs` does `except Exception: continue`, so if a CLAUDE.md / DECISIONS.md / memory file is unreadable (permissions, encoding/UnicodeDecodeError, transient I/O), its paragraphs silently drop out of the dedup corpus. The user-visible effect: an already-captured insight gets re-surfaced as 'Novel' in QUEUE.md with no signal that the corpus was incomplete. Catch the read narrowly and echo a warning so an operator can tell 'no match' from 'never compared'. e.g.:
```python
        try:
            with open(path) as fh:
                text = fh.read()
        except OSError as e:
            print(f'dream: skipping corpus file {path}: {e}', file=sys.stderr)
            continue
```

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+
+
+def _disposition_of(body: str) -> str | None:
+    for name, rx in _DISPOSITIONS:
+        if rx.search(body):
+            return name
+    return None
+
+
+def build_findings(comments: list[dict]) -> list[dict]:
```
devonjones, line 141: 🤖 **Claude Code** (pr-test-analyzer):


test: HIGH (criticality 8) — `build_findings` reconstructs review threads with non-trivial logic (skip reply roots, 'last decisive reply wins', `reopening` handling, per-finding disposition) and has zero coverage. A regression here silently mis-scores every reviewer. Suggested test `test_build_findings_last_decisive_reply_wins`: given a finding comment with a thread of [acknowledged, reopening, wont_fix] replies, assert disposition=='wont_fix' and reopened==True; and `test_build_findings_skips_reply_roots` asserting in_reply_to_id comments never produce a finding.

plugins/dream/scripts/dreamlib/config.py
```diff
@@ -0,0 +1,128 @@
...
+import re
+import subprocess
+
+
+def _run(cmd: list[str], cwd: str | None = None) -> str:
+    try:
+        return subprocess.run(
+            cmd, capture_output=True, text=True, cwd=cwd, timeout=10
+        ).stdout.strip()
+    except Exception:
```
devonjones, line 23: 🤖 **Claude Code** (silent-failure-hunter):


MEDIUM — `_run` swallows every subprocess failure and returns '', masking the root cause of downstream errors. `except Exception: return ''` hides FileNotFoundError (git/gh not installed), CalledProcessError, TimeoutExpired, and PermissionError alike. The empty string then propagates: `repo_slug()` returns '' and `reviews.fetch_review_comments` raises the generic 'could not determine repo (set $DREAM_REPO)' — which is misleading if the real cause was 'gh not on PATH' or a 10s timeout. Catch narrowly and, when a command unexpectedly fails (non-empty stderr / non-FileNotFound), echo it to stderr so the failure is diagnosable. e.g.:
```python
    try:
        r = subprocess.run(cmd, capture_output=True, text=True, cwd=cwd, timeout=10)
        if r.returncode != 0 and r.stderr.strip():
            print(f'dream: {cmd[0]} failed: {r.stderr.strip()[:120]}', file=sys.stderr)
        return r.stdout.strip()
    except (FileNotFoundError, subprocess.TimeoutExpired) as e:
        print(f'dream: {cmd[0]} unavailable: {e}', file=sys.stderr)
        return ''
```

plugins/dream/scripts/dreamlib/cli.py
```diff
@@ -0,0 +1,367 @@
...
+        _echo("no digests yet — run `distill` first")
+        return 1
+    tot = {"denials": 0, "errors": 0, "corrections": 0, "insights": 0}
+    by_target: dict[str, int] = {}
+    print(f"{'session':12} {'branch':28} {'den':>3} {'cor':>3} {'err':>4} {'ins':>3}")
+    for p in paths:
+        with open(p) as fh:
+            d = json.load(fh)
+        st = d.get("stats", {})
+        ins = d.get("model", {}).get("candidate_insights", []) if d.get("model") else []
```
devonjones: 🤖 **Claude Code** (code-simplifier):


simplify: LOW — this line packs a redundant default into a ternary: the `if d.get("model")` guard already short-circuits, so the `{}` default in `d.get("model", {})` can never fire. Collapse to an identical, easier read:

```suggestion
        ins = (d.get("model") or {}).get("candidate_insights", []) if d.get("model") else []
```

Or even simpler, dropping the ternary entirely (behavior-identical for None/absent/dict): `ins = (d.get("model") or {}).get("candidate_insights") or []`.

plugins/dream/scripts/dreamlib/parse.py
```diff
@@ -0,0 +1,195 @@
...
+    type: str  # 'user' | 'assistant'
+    role: str
+    uuid: str
+    parent: str | None
+    ts: str | None
+    # content, normalized:
+    text: str = ""  # human/assistant prose (no thinking, no tool noise)
+    blocks: list = field(default_factory=list)  # raw content blocks
+    is_real_user: bool = False  # a genuine human turn (not harness noise)
+    tool_uses: list = field(default_factory=list)  # [(id, name, input)]
```
devonjones, line 66: 🤖 **Claude Code** (type-design-analyzer):


type: [MEDIUM] Event.tool_uses and Event.tool_results are typed as bare list with tuple structure documented only in comments. The comment '# [(id, name, input)]' is the only invariant guard — callers in distill.py unpack via 'tid, name, inp = ev.tool_uses[-1]', so a malformed element causes a ValueError at a distance. Minimal fix: use list[tuple[str, str, dict]] for tool_uses and list[tuple[str, bool, str]] for tool_results (Python 3.9+ built-in generics, no extra imports). This converts a comment-documented invariant into a type-checked one and enables pyright/mypy to catch misuse.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+    m = _SEVERITY.search(body)
+    if m:
+        return f"P{m.group(1)}"
+    m = _GEMINI_BADGE.search(body)
+    if m:
+        return {"critical": "P1", "high": "P2", "medium": "P2", "low": "P3"}[m.group(1)]
+    return None
+
+
+def _disposition_of(body: str) -> str | None:
```
devonjones, line 134: 🤖 **Claude Code** (pr-test-analyzer):


test: HIGH (criticality 7) — `_disposition_of` checks `_DISPOSITIONS` in priority order, so ordering bugs (e.g. 'fixed' matching before 'won't fix', or 'won't fix' inside an 'addressed: won't fix that part' reply) would silently mis-bucket acceptance. No test pins the precedence. Suggested test `test_disposition_priority_order`: assert 'withdrawn'/'won't fix'/'out of scope' win over 'fixed'/'addressed'/'acknowledged' when both keywords co-occur, and that a bare 'Fixed.' maps to 'fixed' while 'eventually fixed' does NOT (the `^\W*` anchor).

plugins/dream/scripts/dreamlib/parse.py
```diff
@@ -0,0 +1,195 @@
...
+    # collapse whitespace
+    s = re.sub(r"[ \t]+", " ", s)
+    s = re.sub(r"\n{3,}", "\n\n", s).strip()
+    return s
+
+
+@dataclass
+class Event:
+    idx: int
+    type: str  # 'user' | 'assistant'
```
devonjones, line 57: 🤖 **Claude Code** (type-design-analyzer):


type: [LOW] Event.type is a plain str with the constraint '# user | assistant' documented in a comment. This is a comment-only invariant — nothing prevents an invalid value at construction or mutation time, and callers use 'ev.type == "assistant"' literal comparisons throughout distill.py and reviews.py. Minimal fix: add 'from typing import Literal' and annotate as 'type: Literal["user", "assistant"]'. This is a one-line change that makes the constraint machine-checkable and self-documenting without any runtime cost.

plugins/dream/scripts/dreamlib/parse.py
```diff
@@ -0,0 +1,195 @@
...
+                cleaned = clean_user_text(content)
+                ev.text = cleaned
+                noise = (not cleaned) or any(
+                    cleaned.startswith(p) for p in _NOISE_PREFIXES
+                )
+                ev.is_real_user = not noise
+            else:
+                ev.text = content
+
+        # An assistant text turn is "real" prose if it has text blocks.
```
devonjones: 🤖 **Claude Code** (code-simplifier):


simplify: LOW — this comment ('An assistant text turn is "real" prose if it has text blocks.') sits directly above a bare `events.append(ev)` and describes a classification that never happens here: `is_real_user` is only ever set in the string-content user branch above, never for assistant events. The comment misleads a reader into thinking assistant turns get an is_real_user check at this point. Either drop the stray comment or move it to where assistant prose is actually consumed (e.g. distill.py's decision-skeleton extraction).

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+    (2) /tmp full subagent transcripts (ephemeral — captured while present)."""
+    os.makedirs(REVIEW_OUT, exist_ok=True)
+    seen: set[str] = set()
+    if os.path.exists(HARVEST_FILE):
+        with open(HARVEST_FILE) as fh:
+            for line in fh:
+                try:
+                    seen.add(json.loads(line)["task_id"])
+                except Exception:
+                    pass
```
devonjones, line 394: 🤖 **Claude Code** (silent-failure-hunter):


MEDIUM — harvest dedup-load swallows JSON/IO errors per line with bare `except Exception: pass`, silently shrinking the 'seen' set. If harvest.jsonl has a corrupt line (or a schema change drops the `task_id` key → KeyError), those task-ids drop out of `seen` and their findings get re-appended on the next run, breaking the documented 'idempotent by task-id' guarantee with no signal. At minimum narrow to `(json.JSONDecodeError, KeyError)` and count/echo skips so silent corruption is visible:
```python
                except (json.JSONDecodeError, KeyError):
                    bad += 1
            if bad:
                print(f'dream: {bad} unreadable harvest lines skipped', file=sys.stderr)
```

plugins/dream/skills/dream/SKILL.md
```diff
@@ -0,0 +1,134 @@
...
+scheduler. `--permission-mode auto` lets the classifier gate each tool call and
+fail closed when unattended (the pipeline shells out, so stricter modes stall).
+Because the skill needs local session logs + `~/.dream/` state + a reachable
+ollama, it must run on your machine, not a cloud scheduler.
+
+See [`references/SCHEDULING.md`](../../references/SCHEDULING.md) for copy-paste
+setups: **Linux** (systemd user timer — survives logout via `loginctl
+enable-linger`, catches missed runs), **macOS** (launchd LaunchAgent), and a
+portable **cron** fallback. Stagger `dream` and `dream-reviewers` a few minutes
+apart; both land per-run files in `review/pending/` for the SessionStart hook.
```
devonjones, line 135: 🤖 **Claude Code** (dependency-reviewer):


No Prerequisites section: the dream pipeline requires ollama (an external service, not pre-installed) for the model enrichment pass, but neither SKILL.md documents how to install it. Suggested addition: add a Prerequisites section at the end of SKILL.md listing 'ollama — local model server for stage 1b enrichment. Install: https://ollama.ai or `brew install ollama` (macOS) / `curl -fsSL https://ollama.com/install.sh | sh` (Linux). Pull the model: `ollama pull qwen2.5:7b`. Skip with `--no-model` if ollama unavailable.'

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+    with open(os.path.join(REVIEW_OUT, "findings.json"), "w") as fh:
+        json.dump(findings, fh, indent=2)
+    return findings
+
+
+def _rate(num: int, den: int) -> float | None:
+    return round(num / den, 3) if den else None
+
+
+def synth(findings: list[dict]) -> dict:
```
devonjones, line 207: 🤖 **Claude Code** (pr-test-analyzer):


test: HIGH (criticality 7) — `synth` computes the two acceptance metrics (value vs fix-strict) and the FP/wont-fix/unresolved rates that are the entire point of reviewer scoring, with denominators that differ per metric (`resolved` excludes 'unresolved'/'reopened'; FP rate is over total findings). None of this arithmetic is tested. Suggested test `test_synth_acceptance_metrics`: feed a fixed findings list mixing all dispositions and assert acceptance_value, acceptance_fixstrict, false_positive_rate and unresolved_rate match hand-computed values, including the den==0 -> None case from `_rate`.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+    ("withdrawn", re.compile(r"\bwithdrawn\b|validator refuted|\binvalid\b", re.I)),
+    ("wont_fix", re.compile(r"\bwon'?t[\s-]?fix\b|\bwontfix\b", re.I)),
+    ("out_of_scope", re.compile(r"\bout[\s-]?of[\s-]?scope\b", re.I)),
+    ("fixed", re.compile(r"^\W*fixed\b", re.I)),
+    ("addressed", re.compile(r"^\W*addressed\b", re.I)),
+    ("acknowledged", re.compile(r"^\W*(acknowledged|ack\b|good catch)", re.I)),
+]
+
+# Buckets for the two acceptance metrics (see synth()).
+VALUE_ACCEPT = {"fixed", "addressed", "acknowledged", "out_of_scope"}
```
devonjones, line 66: 🤖 **Claude Code** (type-design-analyzer):


type: [MEDIUM] VALUE_ACCEPT, FIXSTRICT_ACCEPT, and REJECT are three parallel sets that must partition all non-'unresolved' dispositions, but there is no invariant ensuring coverage. If a new disposition label is added to _DISPOSITIONS (line 56) but not to VALUE_ACCEPT or REJECT, synth() silently treats it as 'unresolved' in the acceptance/FP rates — producing misleading scorecard metrics without any error. The current set is {fixed, addressed, acknowledged, out_of_scope, wont_fix, withdrawn} which is coherent, but the gap is a silent footgun for future extension. Minimal fix: add an assertion (or a module-level check) that FIXSTRICT_ACCEPT <= VALUE_ACCEPT and VALUE_ACCEPT | REJECT covers all disposition names from _DISPOSITIONS. A one-liner at module load: 'assert {n for n, _ in _DISPOSITIONS} == VALUE_ACCEPT | REJECT | {"acknowledged", "out_of_scope"}' would catch any future divergence at import time.

plugins/dream/hooks/session-start-proposals.sh
```diff
@@ -0,0 +1,53 @@
...
+proj="${CLAUDE_PROJECT_DIR:-$PWD}"
+root="$(git -C "$proj" rev-parse --show-toplevel 2>/dev/null)" || exit 0
+slug="$(basename "$root")"
+home="${DREAM_HOME:-$HOME/.dream/$slug}"
+[ -d "$home" ] || exit 0
+
+# Outstanding = per-run proposal files sitting in review/pending/. The user
+# dismisses a run by moving its file into review/reviewed/ (or deleting it), so
+# this glob only needs the pending/ dir.
+mapfile -t pending < <(ls -1 "$home"/review/pending/*.md 2>/dev/null)
```
devonjones: 🤖 **Claude Code** (dependency-reviewer):


mapfile is bash 4.0+ only and silently fails (or errors) with macOS default /bin/bash (which is bash 3.2 due to GPL licensing). The SessionStart hook is expected to run on macOS (SCHEDULING.md has a macOS section), so this will silently produce no output there — the pending-file alert will never fire. Suggestion: replace 'mapfile -t pending < <(ls -1 ...)' with a portable equivalent: 'pending=(); while IFS= read -r f; do pending+=("$f"); done < <(ls -1 "$home"/review/pending/*.md 2>/dev/null)' — or add a bash version check at the top with a fallback, or note in the hook header that bash 4+ is required and explain how to install it on macOS (e.g. brew install bash).

plugins/dream/scripts/dream
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+# Entry point for the dream pipeline CLI. Pure-stdlib, so system python is fine.
+# Derives the target project from $DREAM_PROJECT_DIR or the current directory.
+DIR="$(cd "$(dirname "$0")" && pwd)"
```
devonjones, line 4: 🤖 **Claude Code** (dependency-reviewer):


Hardcoded /usr/bin/python3 path may not exist on all systems. On macOS with Homebrew Python, pyenv, or conda environments, python3 lives at /usr/local/bin/python3, /opt/homebrew/bin/python3, or a venv path — not /usr/bin/python3 — causing a silent 'No such file or directory' failure. The script comment says 'Pure-stdlib, so system python is fine' but the path assumption may not hold. Suggestion: use 'exec python3 -m dreamlib.cli "$@"' (relies on PATH, which is correct) or add a shebang-style search: 'PYTHON=$(command -v python3 || command -v python)' with a fallback error message. Also add a Prerequisites note in dream/SKILL.md noting Python 3 is required.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+        if len(chunk) < 100:
+            break
+        page += 1
+    with open(cache, "w") as fh:
+        json.dump(all_c, fh)
+    echo(f"fetched {len(all_c)} review comments across {page} pages")
+    return all_c
+
+
+def _reviewer_of(c: dict) -> str | None:
```
devonjones, line 108: 🤖 **Claude Code** (pr-test-analyzer):


test: MEDIUM (criticality 6) — `_reviewer_of` has a deliberate prose-rejection branch (`_VALID_NAME.match` rejecting a finding body captured inside parens, per the comment) plus gemini/cursor login fallbacks. The reject branch is exactly the kind of guard that rots silently. Suggested test `test_reviewer_of_rejects_prose_and_resolves_logins`: an `<!-- Agent: not a real name -->` body returns None, a kebab name returns it, and a comment from login 'gemini-code-assist[bot]' with no signature returns 'gemini'.

plugins/dream/scripts/dreamlib/parse.py
```diff
@@ -0,0 +1,195 @@
...
+    "Caveat:",
+    "[Request interrupted",
+    "This session is being continued",
+    "Your task is to create a detailed summary",
+)
+
+DENIAL_MARKERS = ("doesn't want to proceed", "tool use was rejected")
+
+
+def clean_user_text(s: str) -> str:
```
devonjones, line 40: 🤖 **Claude Code** (pr-test-analyzer):


test: HIGH (criticality 7) — `clean_user_text` + the `is_real_user` gating (lines 170-177) is the load-bearing classifier that decides whether a user turn becomes friction/correction signal downstream; a regression that lets harness noise through (or drops real prose) corrupts every digest, invisibly. No test exercises it. Suggested test `test_clean_user_text_strips_harness_and_flags_real_turns`: a pure `<system-reminder>…</system-reminder>` event yields empty text and is_real_user==False; a 'Caveat:'-prefixed event is flagged noise; genuine prose with a stray `<tag>` survives with is_real_user==True.

plugins/dream/scripts/dreamlib/parse.py
```diff
@@ -0,0 +1,195 @@
...
+    if isinstance(content, list):
+        parts = []
+        for b in content:
+            if isinstance(b, dict) and b.get("type") == "text":
+                parts.append(b.get("text", ""))
+        return "\n".join(parts)
+    return ""
+
+
+def load_session(path: str) -> Session:
```
devonjones, line 96: 🤖 **Claude Code** (pr-test-analyzer):


test: HIGH (criticality 7) — `load_session` is the parser everything else builds on (tool_use/tool_result extraction, tool_names map, pr-link capture, malformed-line skip at line 115). The bare `except: continue` on bad JSON means a regression could silently drop valid events with no failure. Suggested test `test_load_session_extracts_tools_and_survives_bad_lines`: build a small in-memory JSONL with a tool_use, a matching tool_result(is_error), a blank line and a malformed line; assert the malformed line is skipped, both well-formed events parse, tool_names maps id->name, and tool_results carries is_error==True.

plugins/dream/scripts/dreamlib/cli.py
```diff
@@ -0,0 +1,367 @@
+"""Dream stage-1 CLI: distill Claude Code session logs into behavioral digests.
+
+    python -m dream.cli distill            # distill the backlog (cached)
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


docs: HIGH — Module docstring shows `python -m dream.cli distill` but the actual module path is `dreamlib.cli` (the entry script invokes `python3 -m dreamlib.cli`). Anyone copying this example will get a ModuleNotFoundError. Fix: replace `dream.cli` with `dreamlib.cli` on lines 3–6.

plugins/dream/scripts/dreamlib/distill.py
```diff
@@ -0,0 +1,208 @@
...
+
+def _extract_user_correction(denial_text: str) -> str:
+    """Pull the embedded 'the user said:' instruction out of a denial blob."""
+    m = re.search(r"the user said:\s*(.+)", denial_text, re.I | re.S)
+    if m:
+        return _truncate(m.group(1), 400)
+    return ""
+
+
+def heuristic_digest(session: Session) -> dict:
```
devonjones, line 68: 🤖 **Claude Code** (pr-test-analyzer):


test: HIGH (criticality 7) — `heuristic_digest` is the deterministic friction miner (denial vs error vs correction routing, dedupe-with-occurrences, stats tallies) — the 'ore we actually want' per the docstring, and it runs with NO model. Its branch routing is untested. Suggested test `test_heuristic_digest_classifies_and_dedupes_friction`: a Session with one denial tool_result, one is_error result, and two identical correction turns produces friction with kinds {denial,error,correction}, the repeated correction collapsed to occurrences==2, and stats.denials/errors/corrections matching.

plugins/dream/scripts/dreamlib/config.py
```diff
@@ -0,0 +1,128 @@
+"""Runtime configuration — derive everything from the current project context so
+the skill works on any repo, not just wyrd.
+
+- project_log_dir : where Claude Code stores this project's session JSONL
+- dream_home      : per-project derived-data root (`<git-root>/.dream`, gitignored)
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


docs: HIGH — Module docstring says dream_home is `<git-root>/.dream, gitignored` but the actual implementation returns `~/.dream/<slug>` (outside the repo entirely). This is the opposite of what's documented: data does NOT go in the working tree. Fix: replace with `~/.dream/<slug>` (honoring $DREAM_HOME) as described in DESIGN.md.

plugins/dream/scripts/dreamlib/synth.py
```diff
@@ -0,0 +1,218 @@
...
+    for label, ptoks, snippet in paras:
+        cov = len(rep_tokens & ptoks) / len(rep_tokens)
+        if cov > best_cov:
+            best_cov, best = cov, (label, snippet)
+    if best is None or best_cov < 0.30:
+        return None
+    return {"artifact": best[0], "coverage": round(best_cov, 2), "snippet": best[1]}
+
+
+def _suggest_route(members: list[dict], frequency: int, known: dict | None) -> str:
```
devonjones, line 154: 🤖 **Claude Code** (pr-test-analyzer):


test: MEDIUM (criticality 6) — `_suggest_route` encodes the promotion policy: frequency-gated downgrade of single-session claude_md -> memory, and DECISIONS.md-coverage -> product_decision. These thresholds (frequency<2, KNOWN_THRESHOLD) are exactly the kind of business rule that drifts. Suggested test `test_suggest_route_frequency_and_decision_gates`: members tagged claude_md with frequency==1 route to 'memory'; frequency>=2 stays 'claude_md'; a known_match on a DECISIONS.md artifact at coverage>=0.50 routes to 'product_decision'.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+        ["gh", "api", path], capture_output=True, text=True, timeout=60
+    )
+    if out.returncode != 0:
+        raise RuntimeError(f"gh api {path} failed: {out.stderr[:200]}")
+    return json.loads(out.stdout)
+
+
+def fetch_review_comments(echo=lambda *_: None) -> list[dict]:
+    """Paginate every PR review (line) comment in the repo. Cached to one file —
+    closed/merged PR comments are effectively immutable so a refresh only needs
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


docs: HIGH — fetch_review_comments docstring claims the cache refresh 'only needs to pull pages newer than the last fetch (handled by re-running; cheap)' but the code re-fetches ALL pages from page 1 on every run and overwrites the cache file entirely — there is no incremental/since-last-fetch logic. Fix: remove the false incremental-pull claim; the accurate behavior is 're-running re-fetches all pages and overwrites the cache'.

plugins/dream/scripts/dreamlib/config.py
```diff
@@ -0,0 +1,128 @@
...
+    p = os.path.abspath(cwd or project_dir())
+    munged = p.replace("/", "-")
+    return os.path.expanduser(os.path.join("~/.claude/projects", munged))
+
+
+def memory_dir(cwd: str | None = None) -> str:
+    return os.path.join(project_log_dir(cwd), "memory")
+
+
+def repo_slug(cwd: str | None = None) -> str:
```
devonjones, line 73: 🤖 **Claude Code** (pr-test-analyzer):


test: MEDIUM (criticality 5) — `repo_slug`'s git-remote fallback parses owner/name out of a URL with a regex (line 84) covering both SSH (`git@host:owner/name.git`) and HTTPS forms plus the `.git` strip. Easy to regress to a wrong slug, which would point the whole GitHub-mining pipeline at the wrong repo. The `gh`/git `_run` calls can be monkeypatched. Suggested test `test_repo_slug_parses_ssh_and_https_remotes`: assert both `git@github.com:owner/repo.git` and `https://github.com/owner/repo` yield 'owner/repo', and $DREAM_REPO short-circuits both.

plugins/dream/scripts/dreamlib/model.py
```diff
@@ -0,0 +1,126 @@
+"""Stage 1b: enrich a heuristic digest via a local ollama model.
+
+Runs on the macbook box (default http://10.5.2.31:11434). Uses ollama's
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


docs: MEDIUM — Module docstring says 'Runs on the macbook box (default http://10.5.2.31:11434)' — this hardcodes Devon's machine-internal IP as the documented default, which will be wrong for any other user. SCHEDULING.md describes the default as `http://localhost:11434`. Fix: drop 'the macbook box' and say 'default `http://localhost:11434`; override via $WYRD_OLLAMA_URL'.

plugins/pr-review-loop/skills/pr-review-loop/scripts/emit-dream-marker.sh
```diff
@@ -0,0 +1,47 @@
...
+    # only accept identifier-shaped keys (jq --arg requirement)
+    case "$k" in
+      [a-zA-Z_]*) : ;;
+      *) continue ;;
+    esac
+    jqargs+=(--arg "$k" "$v")
+    filter="$filter + {$k:\$$k}"
+  done
+
+  jq -cn "${jqargs[@]}" "$filter" >> "$dir/pr-review-loop.jsonl" 2>/dev/null || return 0
```
devonjones, line 43: 🤖 **Claude Code** (pr-test-analyzer):


test: MEDIUM (criticality 6) — this is one half of a producer/consumer contract: `emit-dream-marker.sh` writes markers that `reviews.read_markers` parses back. The two are in different plugins and nothing pins the JSONL shape, so a key rename on either side breaks reviewer scoring silently (the consumer skips unrecognized markers without erroring). Suggested test `test_emit_then_read_markers_roundtrip`: run this script with a `reviewer-finding` and assert `read_markers()` returns one finding with the expected reviewer/severity/disposition fields. NOTE: the dream plugin currently ships ZERO Python tests — recommend a beads follow-up to stand up a pytest harness rather than blocking this PR on full coverage.

plugins/dream/references/SCHEDULING.md
```diff
@@ -0,0 +1,181 @@
...
+> `~/.dream/<slug>/` state, and — for the model pass — a reachable ollama). So it
+> must run **on the machine where those live**, not in a cloud scheduler.
+
+## The headless command
+
+```bash
+claude -p "Run the dream skill on this project: distill new sessions, synth, then \
+consolidate. Auto-write only genuinely-new memories; collect every proposal into \
+the skill's review/pending/ file; do NOT auto-edit CLAUDE.md, skills, or DECISIONS.md. \
+Print a short summary with a per-bucket count. Do not manufacture insights." \
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


**Verbose:** `Print a short summary with a per-bucket count.` → `Print a per-bucket count.` — "a short summary with" is redundant; the per-bucket count IS the summary. The other prompt variants in this file (lines 59, 121, 166) already use the tighter form.

plugins/dream/scripts/dreamlib/synth.py
```diff
@@ -0,0 +1,218 @@
...
+The DETERMINISTIC mechanical core. It does the cheap, repeatable heavy lifting
+so the expensive Opus judgment pass can be small and focused:
+
+  1. Pool every candidate_insight across all digests.
+  2. Cluster near-duplicate insights (lexical) so the same lesson across N
+     sessions becomes one cluster with frequency=N. Frequency is the promotion
+     signal (a 1-session lesson is a memory at most; a recurring one is a
+     CLAUDE.md candidate).
+  3. Pre-flag "likely already captured" by lexical match against the KNOWN
+     corpus — existing memories, both CLAUDE.md files, and kenning DECISIONS.md.
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


docs: MEDIUM — Docstring says 'known corpus — existing memories, both CLAUDE.md files, and kenning DECISIONS.md' but the implementation is generic: it collects ALL DECISIONS.md under any git root, not just wyrd's kenning file, and gathers all CLAUDE.md/AGENTS.md up the tree. 'kenning DECISIONS.md' is wyrd-project-specific; this is a general-purpose module. Fix: 'and any DECISIONS.md files auto-discovered under the git root'.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -1299,6 +1303,40 @@ During the review loop, if you identify guidance that would improve review quali
...
+
+Rules:
+- **`disposition_by=user`** whenever the operator made the call (overruled a
+  finding, signed off a P1/P2 carry-forward, resolved a self-contradiction). This
+  is the single most valuable signal dream gets — default to `orchestrator`, set
+  `user` only when the human actually weighed in.
+- **Best-effort, never blocking.** The script no-ops silently if `jq`/git are
+  absent; a marker write must never interrupt the loop. Emit it alongside the
+  thread reply, not instead of it — the PR thread is still the system of record.
+- Markers append to `~/.dream/<repo>/markers/pr-review-loop.jsonl` (outside the
```


docs: MEDIUM — Dream Markers section says markers go to `~/.dream/<repo>/markers/` but every other location in the system (MARKER-CONTRACT.md, emit-dream-marker.sh, DESIGN.md) uses `~/.dream/<slug>/` where slug is the git-root basename. Using `<repo>` here creates a naming inconsistency a reader could mistake for a different variable. Fix: replace `<repo>` with `<slug>` to match the rest of the system's terminology.

plugins/dream/skills/dream/SKILL.md
```diff
@@ -0,0 +1,134 @@
...
+fold in any structured markers producers left for you.
+
+## 0. Locate the scripts
+
+Use Glob to find the entry wrapper, then use its full literal path for every call:
+```
+Glob: **/plugins/dream/scripts/dream      (clone)   OR   **/dream/*/scripts/dream  (installed)
+```
+Run everything from the **target project directory** (so paths derive correctly),
+or pass `DREAM_PROJECT_DIR=/path/to/project`. The tool needs nothing wyrd-specific
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


**Unexplained jargon in a public plugin:** `The tool needs nothing wyrd-specific` — "wyrd" is the name of an unrelated private project; readers of this marketplace plugin have no context for it. Suggest: `The tool has no project-specific dependencies` or just delete the clause (the rest of the sentence already says what it derives).

plugins/dream/hooks/session-start-proposals.sh
devonjones: 🤖 **Claude Code** (code-reviewer):


bug: Important (confidence 85) — mapfile is a bash 4.0+ builtin; macOS ships bash 3.2 by default. With set -u active (line 15), if mapfile is unavailable the 'pending' array is never assigned, and ${#pending[@]} on the next line triggers 'unbound variable', causing the script to exit non-zero. This violates the explicit contract in the header: 'always exits 0 and never errors out — a SessionStart hook must never block a session.' Fix: replace mapfile with a POSIX-compatible alternative, e.g. pending=(); while IFS= read -r f; do pending+=("$f"); done < <(ls -1 "$home"/review/pending/*.md 2>/dev/null) — or add a bash version guard and fall back gracefully.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+        with open(path) as fh:
+            for line in fh:
+                line = line.strip()
+                if not line:
+                    continue
+                try:
+                    m = json.loads(line)
+                except Exception:
+                    continue
+                if m.get("kind") not in (None, "reviewer-finding"):
```
devonjones: 🤖 **Claude Code** (code-reviewer):


bug: Important (confidence 80) — read_markers() passes kind=None through its reviewer-finding filter. The MARKER-CONTRACT.md states 'Every record has ts (ISO-8601 UTC), skill (producer), and kind.' Allowing kind=None means a malformed record (or a future record from a different producer that lacks kind but happens to include a 'reviewer' field) is silently classified as a reviewer-finding and counted in the scorecards. The secondary guard (if not m.get('reviewer')) reduces the practical blast radius but doesn't make the logic correct. Fix: change the filter to if m.get('kind') != 'reviewer-finding': continue — accept only explicit reviewer-finding records, consistent with the contract.

plugins/dream/skills/dream/SKILL.md
```diff
@@ -0,0 +1,134 @@
...
+SessionStart hook warns about anything there. The user dismisses a run by moving
+its file into `review/reviewed/` (or deleting it) once actioned — so keep each
+run's proposals in its own file and never write into `review/reviewed/` yourself.
+
+Report what you auto-wrote and what you propose (with a per-bucket count, including
+a `decisions-log` count even when it is 0). Done.
+
+## Running on a schedule (nightly, unattended)
+
+To mine the corpus automatically, run this skill headlessly on a timer:
```
devonjones: 🤖 **Claude Code** (clarity-reviewer):


**Filler opener:** `To mine the corpus automatically, run this skill headlessly on a timer:` — the section heading "Running on a schedule (nightly, unattended)" already establishes this context. Suggest starting with the command directly: `Run this skill headlessly on a timer:` or just omit the leading clause.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
```diff
@@ -1299,6 +1303,40 @@ During the review loop, if you identify guidance that would improve review quali
...
+  repo). Schema: the dream plugin's `references/MARKER-CONTRACT.md`.
+
+This makes the per-flagger acceptance telemetry the validator section refers to
```
devonjones, line 1337: 🤖 **Claude Code** (clarity-reviewer):


**Redundant closing sentence:** `This makes the per-flagger acceptance telemetry the validator section refers to real and durable, instead of something reconstructed from lossy logs after the fact.` — this repeats what lines 1308-1313 already state ("a finding that's fixed without being posted leaves no trace", "threads can't say whether *you* or the orchestrator made the call"). Delete this sentence; the reader already has the rationale.

plugins/dream/hooks/session-start-proposals.sh
devonjones: Fixed in 303dd0c - replaced mapfile (bash 4.0+) with a portable while-read loop so the hook works on macOS bash 3.2 while keeping its never-block contract.

devonjones: Fixed in 303dd0c - replaced the bash-4 mapfile with a portable while-read loop (macOS ships bash 3.2).

plugins/dream/references/SCHEDULING.md
devonjones: Fixed in 303dd0c - tightened to 'Print a per-bucket count.'

plugins/dream/skills/dream/SKILL.md
devonjones: Fixed in 303dd0c - now 'nothing project-specific' (dropped the wyrd reference in this public-plugin doc).

plugins/dream/skills/dream/SKILL.md
devonjones: Fixed in 303dd0c - dropped the filler opener; the section heading already says it.

plugins/dream/skills/dream/SKILL.md
devonjones: Fixed in 303dd0c - added an ollama install pointer (ollama.com) plus a stage-1 env-var reference.

plugins/dream/scripts/dreamlib/config.py
devonjones: Fixed in 303dd0c - dream_home docstring now says ~/.dream/<slug> (outside the repo), matching the code.

plugins/dream/scripts/dreamlib/cli.py
devonjones: Fixed in 303dd0c - docstring examples now use dreamlib.cli (the module the dream wrapper actually runs).

plugins/dream/scripts/dreamlib/cli.py
devonjones: Fixed in 303dd0c - collapsed the redundant {} default; the ternary is already guarded by if d.get('model').

plugins/dream/scripts/dream
devonjones: Fixed in fd1c058 - shebang now '/usr/bin/env python3' for Homebrew/pyenv/conda macOS.

plugins/dream/scripts/dreamlib/synth.py
devonjones: Fixed in fd1c058 - genericized to 'any project DECISIONS.md'.

plugins/dream/scripts/dreamlib/model.py
devonjones: Fixed in fd1c058 - default is now http://localhost:11434 (in model.py and cli.py); the private IP was a wyrd-ism. Override via WYRD_OLLAMA_URL.

plugins/dream/scripts/dreamlib/reviews.py
devonjones: Fixed in fd1c058 - docstring corrected: each call re-fetches all pages (no incremental cursor); the old 'pages newer than last fetch' claim was false.

plugins/dream/scripts/dreamlib/parse.py
devonjones: Fixed in fd1c058 - removed the misleading comment.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Fixed in fd1c058 - <repo> -> <slug> to match every other reference.

plugins/dream/scripts/dreamlib/reviews.py
devonjones: Fixed in fd1c058 - read_markers now requires kind=='reviewer-finding'; per MARKER-CONTRACT every record has a kind, so accepting None could ingest session-miner markers.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+        try:
+            d = json.loads(line)
+        except Exception:
+            continue
+        if d.get("type") != "assistant":
+            continue
+        ts = d.get("timestamp")
+        for b in (d.get("message", {}) or {}).get("content", []) or []:
+            if not isinstance(b, dict) or b.get("type") != "tool_use":
+                continue
```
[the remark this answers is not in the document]
devonjones: Fixed in fd1c058 - added isinstance(message, dict) / isinstance(content, list) guards in coverage_from_logs.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+    tmp_by_id = {}
+    for p in _glob.glob(TMP_TASK_GLOB):
+        tmp_by_id[os.path.basename(p)[:-7]] = p  # strip ".output"
+    tmp_captured = 0
+    for rec in new:
+        path = tmp_by_id.get(rec["task_id"])
+        if path:
+            try:
+                with open(path) as fh:
+                    rec["tmp_transcript_bytes"] = os.path.getsize(path)
```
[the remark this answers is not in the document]
devonjones: Fixed in fd1c058 - dropped the redundant with-open; getsize() needs no handle.

plugins/dream/scripts/harvest-cron.sh
```diff
@@ -0,0 +1,22 @@
...
+# this is best-effort opportunistic capture; the durable, authoritative source
+# going forward is the dream-marker stream (.dream/markers/). Idempotent (dedup
+# by task-id), so running daily is safe and cheap.
+set -euo pipefail
+
+PROJECT="${1:?usage: harvest-cron.sh <project-dir>}"
+HERE="$(cd "$(dirname "$0")" && pwd)"
+cd "$PROJECT"
+# dream_home lives outside the repo at ~/.dream/<slug>; log alongside it.
+STATE="$HOME/.dream/$(basename "$PROJECT")"
```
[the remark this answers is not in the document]
devonjones: Fixed in fd1c058 - uses basename "$PWD" after the cd, so relative project paths resolve correctly.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+def _pr_of(desc: str) -> str | None:
+    m = _PR_IN_DESC.search(desc)
+    if m:
+        return m.group(1) or m.group(2)
+    return None
+
+
+def _iter_log_lines():
+    for f in sorted(_glob.glob(os.path.join(LOGS_DIR, "*.jsonl"))):
+        with open(f) as fh:
```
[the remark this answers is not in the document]
devonjones: Fixed in fd1c058 - encoding=utf-8.

plugins/dream/scripts/dreamlib/reviews.py
```diff
@@ -0,0 +1,504 @@
...
+# --------------------------------------------------------------------------
+
+
+def read_markers() -> list[dict]:
+    """Read the dream-marker stream (.dream/markers/*.jsonl) and normalize each
+    reviewer-finding marker into the finding shape `synth` consumes. See
+    references/MARKER-CONTRACT.md for the schema producers honor."""
+    findings: list[dict] = []
+    for path in sorted(_glob.glob(os.path.join(config.markers_dir(), "*.jsonl"))):
+        with open(path) as fh:
```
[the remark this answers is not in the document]
devonjones: Fixed in fd1c058 - encoding=utf-8.

plugins/dream/scripts/dreamlib/reviews.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/reviews.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/reviews.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/reviews.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/parse.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/parse.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/distill.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/synth.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/config.py
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/pr-review-loop/skills/pr-review-loop/scripts/emit-dream-marker.sh
devonjones: Out of scope for this PR - tracked in devon-claude-skills-5wu (add pytest harness + tests for the dream plugin). Valid coverage gap; standing up the harness is its own PR.

plugins/dream/scripts/dreamlib/synth.py
devonjones: Partially addressed in fd1c058 - added encoding=utf-8 (the likely failure on emoji-laden corpus files). Kept the except-continue deliberately: one unreadable known-corpus file shouldn't abort the whole consolidation. Tradeoff acknowledged.

plugins/dream/hooks/session-start-proposals.sh
devonjones: Won't fix (reclassify P1->P3): set -e would break this hook's contract — a SessionStart hook must ALWAYS exit 0 and never block the session. set -u is kept; explicit exit 0 + guards handle the rest by design.

plugins/pr-review-loop/skills/pr-review-loop/scripts/emit-dream-marker.sh
devonjones: Won't fix (reclassify P1->P3): emit-dream-marker is best-effort and 'must never interrupt the loop' per its contract; set -e would defeat that. It already no-ops silently on missing jq/git.

plugins/dream/scripts/dreamlib/parse.py
devonjones: Acknowledged - internal dataclass; the (id,name,input) tuple shape is documented and stable across the one producer/consumer. Full generic typing here adds noise without catching a real bug.

plugins/dream/scripts/dreamlib/parse.py
devonjones: Acknowledged - a Literal would be marginally nicer, but the two values are documented inline and the field is set in one place. Low value.

plugins/dream/scripts/dreamlib/reviews.py
devonjones: Acknowledged - the disposition sets are a small closed set defined next to _DISPOSITIONS; a coverage assert is reasonable but low-value given how rarely they change.

plugins/dream/scripts/dreamlib/config.py
devonjones: Intentional - _run returns '' so callers degrade gracefully when git/the remote is absent (slug/branch are optional); raising would break non-git or detached checkouts.

plugins/dream/scripts/dreamlib/reviews.py
devonjones: Acknowledged - best-effort dedup load; the harvest append is idempotent by task_id, so a corrupt line at worst re-harvests a single record on the next run.

plugins/pr-review-loop/skills/pr-review-loop/SKILL.md
devonjones: Won't fix - this is a deliberate section-closing summary tying the rationale back to the validator telemetry, not a redundant restatement.

plugins/pr-review-loop/skills/pr-review-loop/scripts/emit-dream-marker.sh
```diff
@@ -0,0 +1,47 @@
...
+  local jqargs=(--arg ts "$ts" --arg skill "pr-review-loop" --arg kind "$kind")
+  local filter='{ts:$ts, skill:$skill, kind:$kind}'
+  local kv k v
+  for kv in "$@"; do
+    k="${kv%%=*}"; v="${kv#*=}"
+    # only accept identifier-shaped keys (jq --arg requirement)
+    case "$k" in
+      [a-zA-Z_]*) : ;;
+      *) continue ;;
+    esac
```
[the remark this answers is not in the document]
devonjones: Fixed - replaced the first-char-only glob with an anchored regex [[ "$k" =~ ^[a-zA-Z_][a-zA-Z0-9_]*$ ]] that rejects hyphens anywhere, so a bad key is skipped rather than silently breaking the jq filter. Verified: bad-key/9bad skip, pr/severity/_ok accept.

plugins/dream/scripts/dreamlib/config.py
```diff
@@ -0,0 +1,110 @@
...
+    d = cwd
+    while True:
+        for name in ("CLAUDE.md", "AGENTS.md"):
+            p = os.path.join(d, name)
+            if os.path.exists(p):
+                files.append(p)
+        if d == root or os.path.dirname(d) == d:
+            break
+        d = os.path.dirname(d)
+    for extra in (os.environ.get("DREAM_EXTRA_CORPUS", "").split(":")):
```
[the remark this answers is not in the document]
devonjones: Fixed - splits DREAM_EXTRA_CORPUS on os.pathsep now.

plugins/dream/skills/dream/SKILL.md
```diff
@@ -0,0 +1,135 @@
...
+## 1. Distill (stage 1 — deterministic + local model)
+
+```bash
+<scripts>/dream distill          # all un-distilled sessions (cached by source hash)
+<scripts>/dream distill --no-model   # heuristic only (fast, skips ollama)
+<scripts>/dream stats            # friction + candidate-insight tallies
+```
+
+Distillation extracts **friction** (denials, corrections, tool errors) and runs a
+local ollama model (`WYRD_OLLAMA_URL`, default `http://10.5.2.31:11434`;
```
devonjones: 🤖 **Claude Code** (comment-analyzer):


docs: HIGH — SKILL.md line 47 still documents the old private IP (http://10.5.2.31:11434) as the WYRD_OLLAMA_URL default, but fix commits 303dd0c/fd1c058 updated the actual default in model.py and cli.py to http://localhost:11434. Anyone reading the docs gets the wrong default. Suggested fix: replace `http://10.5.2.31:11434` with `http://localhost:11434` on this line.

devonjones: Fixed - SKILL.md:47 now says http://localhost:11434, consistent with the model.py/cli.py defaults. Good catch; this was an oversight in the genericization fix.

